PR #1, opened 2020-05-31
implement github actions for helm charts

billimek:
Inspired by [this example repo](https://github.com/helm/chart-testing-action), this PR should implement GitHub Actions support which will:

* For every PR impacting the helm chart, a linter will make sure that the change passes linting which includes forcing a version bump on changes)
* For every PR impacting the helm chart, a 'kubernetes in docker' (KIND) cluster will be temporarily created and the chart will be installed to that test cluster to ensure that it installs properly
* For every merge to master impacting the helm chart, automation will package the chart, update the `index.yaml` on the `gh-pages` branch, and cut a github release with the packaged chart for reference as a chart repository.

**ACTION REQUIRED:**

* Github Pages should be enabled within the repo settings. Please use the default `gh-pages` branch
* Create a gtihub personal access token with `repo` permissions scope, make note of the token
* Create a new repo secret (accessible within the repo settings) named `CR_TOKEN` with the value being the github access token created above

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,46 @@
+
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Turnstyle
+        uses: softprops/turnstyle@v1
+        with:
+          continue-after-seconds: 180
+        env:
+          GITHUB_TOKEN: ${{ secrets.CR_TOKEN }}
+
+      - name: Fetch history
+        run: git fetch --prune --unshallow
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      # See https://github.com/helm/chart-releaser-action/issues/6
+      - name: Install Helm
+        run: |
+          curl -fsSLo get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
+          chmod 700 get_helm.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Turnstyle
+        uses: softprops/turnstyle@v1
+        with:
+          continue-after-seconds: 180
+        env:
+          GITHUB_TOKEN: ${{ secrets.CR_TOKEN }}
```
itzg, line 22: I created `CR_TOKEN` but just curious if `secrets.GITHUB_TOKEN` could be used here.

billimek: I believe it should be fine to use the `secrets.GITHUB_TOKEN` instead, I'll update the PR to reflect.  The `CR_TOKEN` reference comes from the [implementation of the chart releaser action](https://github.com/helm/chart-releaser-action/blob/master/cr.sh#L23) and must be set as a local environment variable named that in the action environment setting.

This does not mean, of course, that the _GitHub Secret_ must also be named the same.

billimek: This change should now be present!

billimek:
> Glad that works! So, it dawned on me that the actual chart content needs to moved over. How do you recommend I/we proceed with that?

My suggestion is to:

* bring the chart over as a separate PR
* add this repo to helm hub as a [PR over there](https://github.com/helm/hub#how-to-add-your-helm-charts)
* flag the [stable/charts minecraft chart](https://github.com/helm/charts/tree/master/stable/minecraft) as deprecated with appropriate verbiage pointing to this repo in that chart's README (as a PR over there of course)

itzg:
If you don't mind taking the first bullet to give it proper verification, etc
> bring the chart over as a separate PR

then I'm happy to knock out the other two

PR #2, opened 2020-06-09
adding minecraft chart

billimek:
Importing 'minecraft' chart from helm/charts to this repo.

* Bumped chart version to 2.0.0 so that it will appear greater than the version on helm/charts (this will help when using `helm search`)
* Naming the chart `minecraft` to maintain compatibility with existing chart, but we can give it a different name if that makes more sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,26 @@
 # minecraft-server-charts
+
+[![](https://github.com/itzg/minecraft-server-charts/workflows/Release%20Charts/badge.svg?branch=master)](https://github.com/itzg/minecraft-server-charts/actions)
+
+## Usage
+
+[Helm](https://helm.sh) must be installed to use the charts.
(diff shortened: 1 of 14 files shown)


billimek:
@itzg here is the minecraft chart imported from the helm/charts repo.

There are a couple of differences from the upstream chart and this one:

* This chart defaults to ClusterIP instead of LoadBalancer.  This is simply because the chart testing infrastructure doesn't handle LoadBalancer, which is not an issue in the testing infrastructure currently used by the helm/charts repo (because they use google GKE clusters which has the ability to do LoadBalancer services by default)
* This chart defaults to no data persistence.  For reasons I cannot understand, the default `kind` cluster isn't properly provisioning storage during the chart test installation.

billimek:
FYI this is an example pull request I raised to 'deprecate' the node-red chart on the helm/charts repo: https://github.com/helm/charts/pull/22739

PR #4, opened 2020-06-25
Init bedrock chart

EStork09:
Adding bedrock Chart
There are a few things that should probably get updated. I believe there might be conflicts if you ran the minecraft and minecraft-bedrock charts in the same namespace.

### DIFF
--- a/charts/minecraft-bedrock/README.md
+++ b/charts/minecraft-bedrock/README.md
@@ -64,16 +64,17 @@ you can change the values.yaml to disable persistence under the sub-sections und
 
 > *"An emptyDir volume is first created when a Pod is assigned to a Node, and exists as long as that Pod is running on that node. When a Pod is removed from a node for any reason, the data in the emptyDir is deleted forever."*
 
-## Backups - WIP
+## Backups
 
 You can backup the state of your minecraft server to your local machine via the `kubectl cp` command.  
 
 ```shell
 NAMESPACE=default
 POD_ID=lionhope-387ff8d-sdis9
-kubectl attach --namespace ${NAMESPACE} ${POD_ID} 
+kubectl attach --namespace ${NAMESPACE} ${POD_ID} -it
 save hold
 save query
+save resume
+^P + ^Q (CtrlP and CtrlQ)
 kubectl cp ${NAMESPACE}/${POD_ID}:/data .
-kubectl attach --namespace ${NAMESPACE} ${POD_ID} save resume
 ```
--- a/charts/minecraft-bedrock/README.md
+++ b/charts/minecraft-bedrock/README.md
@@ -1,0 +1,78 @@
+# Minecraft
+
+[Minecraft](https://minecraft.net/en/) is a game about placing blocks and going on adventures.
+
+## Introduction
+
+This chart creates a single Minecraft Pod, plus Services for the Minecraft server and RCON.
+
+## Prerequisites
+
+- 512 MB of RAM
+- Kubernetes 1.4+ with Beta APIs enabled
+- PV provisioner support in the underlying infrastructure
+
+## Installing the Chart
+
+To install the chart with the release name `minecraft`, read the [Minecraft EULA](https://account.mojang.com/documents/minecraft_eula) run:
+
+```shell
+helm install minecraft-bedrock \
+  --set minecraftServer.eula=true itzg/minecraft
+```
+
+This command deploys a Minecraft dedicated server with sensible defaults.
+
+> **Tip**: List all releases using `helm list`
+
+## Uninstalling the Chart
+
+To uninstall/delete the `minecraft-bedrock` deployment:
+
+```shell
+helm delete minecraft-bedrock
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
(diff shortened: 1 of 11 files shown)


charts/minecraft-bedrock/templates/deployment.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - name: PLAYER_IDLE_TIMEOUT
+          value: {{ .Values.minecraftServer.playerIdleTimeout | quote }}
+        - name: MAX_THREADS
+          value: {{ .Values.minecraftServer.maxThreads | quote }}
+        - name: GAMEMODE
+          value: {{ .Values.minecraftServer.gameMode | quote }}
+        - name: LEVEL_TYPE
+          value: {{ .Values.minecraftServer.levelType | quote }}
+        - name: GENERATOR_SETTINGS
+          value: {{ default "" .Values.minecraftServer.generatorSettings | quote }}
```
itzg, line 86: Looks like this got accidentally left here from the java edition copy

charts/minecraft-bedrock/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+You can backup the state of your minecraft server to your local machine via the `kubectl cp` command.  
+
+```shell
+NAMESPACE=default
+POD_ID=lionhope-387ff8d-sdis9
+kubectl exec --namespace ${NAMESPACE} ${POD_ID} rcon-cli save-off
+kubectl exec --namespace ${NAMESPACE} ${POD_ID} rcon-cli save-all
+kubectl cp ${NAMESPACE}/${POD_ID}:/data .
+kubectl exec --namespace ${NAMESPACE} ${POD_ID} rcon-cli save-on
```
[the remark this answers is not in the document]
EStork09:
```shell
NAMESPACE=default
POD_ID=lionhope-387ff8d-sdis9
kubectl attach --namespace ${NAMESPACE} ${POD_ID} 
save hold
save query
kubectl cp ${NAMESPACE}/${POD_ID}:/data .
kubectl attach --namespace ${NAMESPACE} ${POD_ID} save resume
```

This is how I would image it would go, but it seems I don't get a console prompt to even be able to run a command. Even running this just stand alone on docker and following the steps you mentioned [here](https://github.com/itzg/docker-minecraft-bedrock-server#executing-server-commands) (I am not an expert at minecraft servers so I might just be missing something) I am unable to actually see any changes on the server or get any response.

Again, I am not an expert, but could we take advantage of pvc snapshotting or something similar to back up the world or does minecraft require these steps to properly back up the world?

charts/minecraft-bedrock/templates/deployment.yaml
```diff
@@ -0,0 +1,136 @@
...
+      annotations:
+        {{- range $key, $value := .Values.podAnnotations }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
+      {{- end }}
+    spec:
+      containers:
+      - name: {{ template "minecraft.fullname" . }}
+        image: "{{ .Values.image }}:{{ .Values.imageTag }}"
+        imagePullPolicy: Always
```
itzg, line 37: Go ahead and add
```
stdin: true
tty: true
```

to the container, which will solve the `docker attach` not working

charts/minecraft-bedrock/README.md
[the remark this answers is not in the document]
itzg: With a majority of stateful workloads, especially databases, there is a slight risk of snapshotting inconsistent file state. As such, there is usually a method to flush-and-pause file operations or export backups in a consistent manner. The bedrock server appears to do a version of the latter for backups to ensure that the world data files are consistent during a backup/snapshot.

The comment I added about `stdin` and `tty` will resolve the issue you saw with attaching.

charts/minecraft-bedrock/templates/deployment.yaml
EStork09: Ah, I have never had to do that before. Thanks, updated to reflected that and successfully tested it out

charts/minecraft-bedrock/ci/test-values.yaml
```diff
@@ -0,0 +1,3 @@
+# This must be overridden, since we can't accept this for the user.
+minecraftServer.eula: "TRUE"
```
billimek, line 2: This needs to be,

```yaml
minecraftServer:
  eula: "TRUE"
```

For the test helm install to consume it properly.  The existing minecraft chart needs to be fixed as well.

EStork09: I thought that was done intentionally to prevent people from blindly accepting the EULA?

itzg: I’m confused too @billimek. I’d prefer to fix the testing action rather than have the chart come pre-baked with the eula accepted.

itzg: ...oh, wait that’s the test-values. That seems fine then.

billimek: Yeah, this is only present for automated tests and is never used by real users at runtime.

billimek:
This LGTM (for what it's worth!)

PR #5, opened 2020-06-30
[minecraft] Fix EULA setting for CI tests

billimek:
This change only affects CI tests and not actual user installations.


billimek:
@itzg for your consideration ^^ to bring into parity with the minecraft-bedrock change.

PR #6, opened 2020-06-30
[minecraft-bedrock] adding additional bedrock server options

billimek:
* Update repo README to show new chart
* add missing bedrock server options
* using default server values (for things like level name, world name, max players) sourced from [here](https://minecraft.gamepedia.com/Server.properties#Bedrock_Edition_3)


billimek:
@itzg for your consideration.  Some of the changes to default values of levelName and serveName, you may not want. I tried to make them match the 'defaults' presented in [server.properties wiki page](https://minecraft.gamepedia.com/Server.properties#Bedrock_Edition_3).

PR #7, opened 2020-07-02
feat(minecraft): Configurable Minecraft Server Annotations

jstewart612:


jstewart612:
@itzg I ran into a use case where I needed service annotations but couldn't add them through this Helm chart, so I figured I'd submit this PR to add a configurable parameter for it.  Can you take a look and let me know if there are any concerns?

jstewart612:
@itzg Add configurable service annotations to the Minecraft helm chart.  Any objections?

PR #9, opened 2020-07-10
[minecraft] - Remove deprecated storageclass annotation

billimek:
Using annotations for storageclass names was [deprecated in kubernetes
1.6](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#using-dynamic-provisioning) and causes issues with some backup/restore scenarios when still present.

This should still operate properly without the old-style annotation approach.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 2.0.2
+version: 2.0.3
 appVersion: 1.14.4
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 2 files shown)


billimek:
@itzg I wonder if we should roll-back this change.  It's apparently a breaking change with existing Minecraft installations because the PVC annotation field is immutable, so the helm upgrade fails.

Thew way I see it, the only way forward is to either:

* treat this as a breaking change (which requires a removal and deletion of the chart followed by a fresh installation), or
* we roll-back this change and live with the deprecated annotation in the PVC

Any thoughts?

itzg:
Yeah, I didn't think about that impact either. It does seem like we should roll-back for now and think if there's some creative path to redo it.

PR #10, opened 2020-07-10
[minecraft-bedrock] - Remove deprecated storageclass annotation

billimek:
Using annotations for storageclass names was [deprecated in kubernetes
1.6](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#using-dynamic-provisioning) and causes issues with some backup/restore scenarios when still present.

This should still operate properly without the old-style annotation approach.

Also needed to bump the server version to '1.16.1.02' in order for it to download properly from upstream

### DIFF
--- a/charts/minecraft-bedrock/Chart.yaml
+++ b/charts/minecraft-bedrock/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-bedrock
-version: 1.1.0
+version: 1.1.1
 appVersion: 1.14.4
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft-bedrock/Chart.yaml
+++ b/charts/minecraft-bedrock/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: minecraft-bedrock
 version: 1.1.1
-appVersion: 1.14.4
+appVersion: 1.16.1
 home: https://minecraft.net/
 description: Minecraft server
 keywords:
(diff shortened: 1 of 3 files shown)


charts/minecraft-bedrock/values.yaml
```diff
@@ -29,7 +29,7 @@ minecraftServer:
   # This must be overridden, since we can't accept this for the user.
   eula: "FALSE"
   # One of: LATEST, SNAPSHOT, or a specific version (ie: "1.7.9").
-  version: "1.14.4"
+  version: "1.16.1.02"
```
itzg, line 32: Actually this would be a good default:
```suggestion
  version: "LATEST"
```

charts/minecraft-bedrock/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: minecraft-bedrock
-version: 1.1.0
-appVersion: 1.14.4
+version: 1.1.1
+appVersion: 1.16.1
```
itzg, line 4: appVersion is always a strange animal IMHO. Perhaps just `1.16` to track it fairly close

charts/minecraft-bedrock/values.yaml
billimek: Agreed, will change this!

charts/minecraft-bedrock/Chart.yaml
billimek: Works for me, will change this too!

billimek:
@itzg I wonder if we should roll-back this change.  It's apparently a breaking change with existing Minecraft installations because the PVC annotation field is immutable, so the helm upgrade fails.

Thew way I see it, the only way forward is to either:

* treat this as a breaking change (which requires a removal and deletion of the chart followed by a fresh installation), or
* we roll-back this change and live with the deprecated annotation in the PVC

Any thoughts?

PR #11, opened 2020-07-13
[minecraft] - Revert #9

billimek:
Reverts itzg/minecraft-server-charts#9


charts/minecraft/templates/datadir-pvc.yaml
```diff
@@ -9,6 +9,11 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
   annotations:
+  {{- if .Values.persistence.storageClass }}
```
itzg, line 12: Throwing out an idea, perhaps users could opt-into the non-annotation based storageClass via a new value condition here.

PR #14, opened 2020-07-15
Fixes #13 Default to LATEST minecraftServer

### LINKED ISSUE #13: Default version update?
I petition that we should release a new chart version with an updated app version 1.16.1 is out, and we are still defaulting to 1.14.4. I am happy to submit a PR, but before I do, I wanted to see if there was some reason why we were still defaulting to 1.14.4.

TJM:


charts/minecraft/Chart.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: v1
 name: minecraft
 version: 2.0.4
```
itzg, line 3: Bump the patch version, please:
```suggestion
version: 2.0.5
```

TJM: 👍

TJM:
The linter is incorrect...

https://v2.helm.sh/docs/developing_charts/#the-appversion-field

That field is optional.

itzg:
> The linter is incorrect...
> 
> https://v2.helm.sh/docs/developing_charts/#the-appversion-field
> 
> That field is optional.

Yeah, I'm not a big fan of `appVersion` in a lot of cases. Hmm, I'll research the linter, but @billimek do you know off hand how to relax that aspect?

TJM:
I worked around it, cause it says that is not used for anything, just informational :)

charts/minecraft/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: minecraft
-version: 2.0.4
-appVersion: 1.14.4
+version: 2.0.5
+appVersion: SeeValues
```
itzg, line 4: That works for me. @billimek , have you seen a convention out there for this?

billimek:
The linter is incorrect according to the [helm docs](https://helm.sh/docs/topics/charts/#the-appversion-field) (@TJM you linked docs from helm v2 but we're using v3 FYI but it's still the same):
image

I'll PR a change to the linter to flag it optional (ref https://github.com/helm/chart-testing/blob/master/etc/chart_schema.yaml#L5).

I looked through all 280+ charts in the helm/stable repo and every single with an `appVersion` field were set to some sort of numeric value.  With that said, I believe using `SeeValues` is still a good path for this change.  The 'appVersion' field is used when you run `helm search` or helm hub type sites, for example:
image

billimek:
FYI https://github.com/helm/chart-testing/pull/239 raised for the linter.

PR #17, opened 2020-09-02
Add externalIPs support for the minecraft service

KristopherKram:
This change will allow users to specify an minecraftServer.externalIPs to allow access to the minecraft service instead of using a load balancer. 

Example:
```Bash
helm upgrade minecraft ./charts/minecraft -n default --set minecraftServer.eula=true --set minecraftServer.externalIPs='{192.168.1.25}'
```


itzg (approved):
Thanks!

PR #19, opened 2020-10-24
Fix capitalization of FORCE_GAMEMODE environment variable

bmxguy100:
The previous capitalization caused `forcegameMode`'s value to be silently ignored


itzg (approved):
Good catch. Thanks!

Issue #20, opened 2020-10-25
whitelisting is very unclear


hyperbolic2346:
The values.yaml says to have a comma separated list of usernames, but the docker container seems to expect true/false with whiltelist in whitelist.json. The chart implies it will turn on whitelisting and generate this json file, but that doesn't seem to be the case. I would suggest either to update the docs to indicate how to turn on whitelisting and mention the json file needs to be filled out or to make the chart write the whitelist.json on start with an init container or the like.

itzg:
The whitelist variable itself is referring to https://github.com/itzg/docker-minecraft-server#whitelist-players , which does end up with the json file getting updated at server startup. With that said, I haven't double checked that it works lately, but I do know the `OPS` setting works correctly in the same way.

hyperbolic2346:
Something is wrong then. When I set `whitelist: player1,player2` I get this:

```
$ kubectl -n minecraft logs minecraft-bedrock-minecraft-bedrock-9ff8555d6-s7swv
DEBU[0000] Using /data to match uid and gid
DEBU[0000] Resolved UID=0 from match path
DEBU[0000] Resolved GID=0 from match path
2020/10/25 14:38:19 Failed to bulk-set properties in file: Value 'player1,player2' for property white-list is in allowed list: [true false]
```

hyperbolic2346:
Oh, I just noticed you linked minecraft and I am using bedrock. I wonder if that is the disconnect here. That appears to pass directly through to the server.

itzg:
Ah, I wonder if the contributor of the bedrock chart accidentally copied that option over as-is. I see what you mean now that it is supposed to be a boolean option in bedrock.

hyperbolic2346:
Yeah, I'm working on a PR to port over the java edition code to build the whitelist.

hyperbolic2346:
Opened https://github.com/itzg/docker-minecraft-bedrock-server/pull/101 to make this behave as it is described in the values.yaml

PR #21, opened 2020-10-26
Update values.yaml

hyperbolic2346:
Adding some detail to the whitelist option.

partially fixes https://github.com/itzg/minecraft-server-charts/issues/20

### DIFF
--- a/charts/minecraft-bedrock/values.yaml
+++ b/charts/minecraft-bedrock/values.yaml
@@ -32,7 +32,8 @@ minecraftServer:
   version: "LATEST"
   # One of: peaceful, easy, normal, and hard
   difficulty: easy
-  # A comma-separated list of player names to whitelist.
+  # A comma-separated list of player names to whitelist with no whitespace.
+  # ex: whitelist: player1,player2,player3
   whitelist:
   # Max connected players.
   maxPlayers: 10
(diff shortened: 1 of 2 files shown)


billimek (approved):

charts/minecraft-bedrock/values.yaml
```diff
@@ -32,7 +32,8 @@ minecraftServer:
   version: "LATEST"
   # One of: peaceful, easy, normal, and hard
   difficulty: easy
-  # A comma-separated list of player names to whitelist.
+  # A comma-separated list of player names to whitelist with no whitespace.
```
itzg, line 35: Isn't the whitelist file also XUIDs like the permissions file?

hyperbolic2346: You can use xuid's or names in the whitelist and the server will fill in the missing data. I'd like to require xuid's like the permissions, but whitelist on java is name-based, so I wanted to keep it the same. I wonder if we should change it in some way to allow both.

itzg: Since names work, then that seems sufficient. Really it's a shame the permissions file doesn't work the same way with filling in the xuid. One of those indicators the bedrock server just isn't fully baked yet.

PR #25, opened 2020-11-02
add prometheus exporter for minecraft

CasparChou:


CasparChou:
@itzg don't know how to fix lint-test failed... is there any more detail logs can check ?

itzg:
Great contribution; however, I would prefer that it incorporate 

https://github.com/itzg/mc-monitor

using its Prometheus exporter mode since that will work for all server types without any additional plugin.

CasparChou:
mc-monitor seems cool !
maybe I can try to add this to chart

itzg:
Let me know if it looks like it'll be too hard to add mc-monitor and the changes you have will be an excellent step in that direction.

CasparChou:
do you prefer to add a chart `mc-monitor` inside this repo, or just add a side-car beside the minecraft-server container ?

itzg:
Since mc-monitor doesn't know how to do service discovery, then deploying as a side car makes more sense.

CasparChou:
itzg/mc-monitor crash after pod started, but it works when I run the code by `go run`, still troubleshooting

itzg:
Oh yeah, that might be an edge case that the code isn't handling well. I'll try to look at that today. 

If I can get to that soon we can go with your original plan and then use a separate PR to add mc-monitor.

itzg:
Even IntelliJ is warning me just by it looking at the code :) 

image

CasparChou:
lol. glad to find out the potential error

itzg:
[0.6.1](https://github.com/itzg/mc-monitor/releases/tag/0.6.1) now includes the fix

ChipWolf:
@CasparChou are you wanting to continue with this? Happy to take over if not

caseyjmorton:
FYI to all interested in this, I'm going to resume work on this as part of my work on a minecraft-operator.  Most of the operator code is based on these helm charts, so I'm happy to do any maintenance needed.  That said, since this PR is almost 3 years old, I'm sure there has been plenty of drift and rather than try to resolve the conflicts, I'm just going to submit a totally new PR.  Once filed, I will post a link to the new one here, for posterity.

PR #27, opened 2020-11-11
Add ability to specify node port for minecraft service

FergusonSean:
This adds the ability to set the nodePort for the service that way we can get a predictable port when deploying this via nodeport. I added some documentation to the values files, but am not sure if this should be documented elsewhere


itzg (approved):
Great idea. Thanks for contributing.

PR #28, opened 2020-11-18
README reflecting current default

roolo:
The default seems to be to disable the persistence — https://github.com/itzg/minecraft-server-charts/blob/d53bc8f12a548a6f7829137b7285b839c5848f22/charts/minecraft/values.yaml#L171


itzg (approved):
Oh yeah, thanks for catching and fixing that.

PR #33, opened 2020-11-29
Fix ENABLE_RCON for minecraftServer.rcon.enable false

### LINKED ISSUE #16: enable-rcon set to true even when RCON is disabled by the chart
Basically what the title says - even if `minecraftServer.rcon.enabled` is `false` when first deployed, the `server.properties` file still has `enable-rcon` set to `true` (at least with the multiarch image.)

MoergJ:
fixes  #16

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 2.0.11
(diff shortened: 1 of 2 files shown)


itzg (commented):
Thanks! Bump the chart version and this is good to go.

itzg (approved):

PR #36, opened 2020-12-09
Adding startup probe

hyperbolic2346:
Adding a startup probe to the chart.

I found that the initial downloads can take a long time. I had a startup world that was about 10 gigs of data, that took time to install as did paper. Adding this startup probe allowed the pod to complete the work before being killed and allows liveness checks to still be peppy to catch crashes.


itzg (approved):
Thanks! This is a great use of startup probes.

Issue #37, opened 2020-12-09
Extra port definitions


hyperbolic2346:
I'm trying to get dynmap to come through with this chart and I'm trying to figure out which way it should be handled. I can see a specific `dynmap: true` flag that does the things or something more generic like `extraPorts` that pastes anything inside into the deployment and service yaml. I can see value in the former because it is easy for people to add dynmap then and the latter because it's easier to select what ports you may need and how to expose them.

Thoughts?

itzg:
Yes, it's hard to decide for the same reasons you gave. I think I'm leaning towards the generic `extraPorts` type approach just in case users come across any other plugins that need others.

hyperbolic2346:
Would you expect these to tag along with the minecraft service as more ports or as services of their own? I originally made this as another port on the service, but I can surely see the desire to keep them separate like the rcon-service.

If we push it to a stand-alone service, it seems more involved than just "we will insert whatever yaml you want into the ports" and more in the realm of "give me a list of ports with defined options and I will build services for you". I'm not sure how far this chart wants to go into yaml config vs options and also just how involved we want this to become.

My immediate desire is to allow dynmap easily in my deployment. I've even toyed with the idea of exposing it via ingress, but that seems overly specific. Maybe dynmap is popular enough that incorporating a dynmap-specific configuration would make sense though?

itzg:
That's a good point about the rcon service being distinct already. Agreed, dynmap does seem to be popular enough to warrant a specific option for it and so far is the only port-exposing plugin I have heard about it. 

Now I'm leaning towards the idea of a dynmap specific option.

itzg:
Fixed in #38

PR #38, opened 2020-12-11
adding dynmap support

hyperbolic2346:
Added dynmap support as discussed in #37. This is more tailored at dynmap and allows easily opening a port or adding an ingress for dynmap, but will not install the dynmap plugin for you.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 2.0.14
+version: 2.1.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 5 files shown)


charts/minecraft/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 2.0.13
+version: 2.0.14
```
billimek, line 3: As this is adding a new feature, can you bump the chart minor version?  In this case, it looks like it would be `2.1.0`.

billimek (commented):
Just one request regarding the version number.

billimek (approved):
LGTM @itzg

itzg (approved):
Great addition! Thanks.

PR #43, opened 2021-01-17
Added serviceAnnotations to minecraft-svc manifests

### LINKED ISSUE #41: minecraft.svc misses annotation
as per the title, minecraft.svc misses a piece, that is contained in the other services
```
   annotations:
     {{ toYaml .Values.minecraftServer.serviceAnnotations | indent 4 }}
```

itzg:
The java edition chart [seems to have the `serviceAnnotations` misplaced](https://github.com/itzg/minecraft-server-charts/blob/4256283a24302ac70d9edd8f7a80b6ea17ad9160/charts/minecraft/values.yaml#L160) compared to the pod and deployment annotations. So, in this PR I'm proposing that `serviceAnnotations` be a top level value.

Went ahead and made the same change for bedrock since it was lacking that value all together.

Fixes #41

### DIFF
--- a/charts/minecraft/templates/rcon-svc.yaml
+++ b/charts/minecraft/templates/rcon-svc.yaml
@@ -5,6 +5,7 @@ metadata:
   name: "{{ template "minecraft.fullname" . }}-rcon"
   annotations:
     {{ toYaml .Values.minecraftServer.serviceAnnotations | indent 4 }}
+    {{ toYaml .Values.serviceAnnotations | indent 4 }}
   labels:
     app: {{ template "minecraft.fullname" . }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
--- a/charts/minecraft-bedrock/Chart.yaml
+++ b/charts/minecraft-bedrock/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-bedrock
-version: 1.1.7
+version: 1.1.8
 appVersion: 1.16
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft-bedrock/templates/minecraft-svc.yaml
+++ b/charts/minecraft-bedrock/templates/minecraft-svc.yaml
@@ -7,6 +7,8 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  annotations:
+    {{ toYaml .Values.serviceAnnotations | indent 4 }}
 spec:
 {{- if (or (eq .Values.minecraftServer.serviceType "ClusterIP") (empty .Values.minecraftServer.serviceType)) }}
   type: ClusterIP
--- a/charts/minecraft-bedrock/values.yaml
+++ b/charts/minecraft-bedrock/values.yaml
@@ -139,3 +139,5 @@ persistence:
 podAnnotations: {}
 
 deploymentAnnotations: {}
+
+serviceAnnotations: {}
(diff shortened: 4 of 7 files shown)


itzg:
cc @srgvg

charts/minecraft/templates/rcon-svc.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: "{{ template "minecraft.fullname" . }}-rcon"
   annotations:
     {{ toYaml .Values.minecraftServer.serviceAnnotations | indent 4 }}
+    {{ toYaml .Values.serviceAnnotations | indent 4 }}
```
billimek, line 8: Do you think we will need a unique `serviceAnnotation` for both the regular service as well as the rcon service, or is combining them ok?

billimek:
Oops, I accidentally merged instead of submitting the 'review' - I think the PR is ok FWIW!

charts/minecraft/templates/rcon-svc.yaml
itzg: Yeah, I'm not sure either way.

srgvg: FWIW, I use this to add a metallb annotation to allow those multiple services to share the same LoadBalancer application. I can't give any insight on possible use case to set this separately, but I'd assume there are, and allowing a separate config might be the the most future proof option?

billimek:
Have either of you upgraded to this version yet, and have rcon enabled?  I just tried to upgrade and got an error, `Helm upgrade failed: YAML parse error on minecraft/templates/rcon-svc.yaml: error converting YAML to JSON: yaml: line 6: did not find expected key`.  Disabling rcon fixes this, so I suspect it something related to the change in `charts/minecraft/templates/rcon-svc.yaml` but don't see how.

Still digging!

itzg:
> Have either of you upgraded to this version yet, and have rcon enabled?

I haven't yet. Maybe it's that awkward "fallback" is did there with the two service annotations blocks. Perhaps should just bite the bullet and make that be just an rcon specific one.

srgvg:
Only got now to test this upgrade, and I'm getting the same error.

I digged a bit further, and noticed I had no problem deploying it from the cloned repo.

So I tested with these both statements:

```
helm upgrade minecraft -f values.yaml --namespace minecraft ./minecraft-server-charts/charts/minecraft/
helm upgrade minecraft -f values.yaml --namespace minecraft itzg/minecraft
```
./minecraft-server-charts/charts/minecraft/ is a clone of this repo at c1f6139

I pulled the charts from the repo:
```
helm pull itzg/minecraft
tar -xvzf minecraft-2.1.2.tgz
```
which unpacked in `./minecraft`

And looking at the diff, I get:
```
--- minecraft-server-charts/charts/minecraft/templates/rcon-svc.yaml	2021-01-26 09:29:13.758224263 +0100
+++ minecraft/templates/rcon-svc.yaml	2021-01-19 03:33:06.000000000 +0100
@@ -4,7 +4,8 @@
 metadata:
   name: "{{ template "minecraft.fullname" . }}-rcon"
   annotations:
-{{ toYaml .Values.rconServiceAnnotations | indent 4 }}
+    {{ toYaml .Values.minecraftServer.serviceAnnotations | indent 4 }}
+    {{ toYaml .Values.serviceAnnotations | indent 4 }}
   labels:
     app: {{ template "minecraft.fullname" . }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
Looking further, the tag curently linked to `master` in this repo is `minecraft-2.1.4`, but I got `2.1.2`.
I checked https://itzg.github.io/minecraft-server-charts/index.yaml, which did have `2.1.4`.

And then it dawned to me.

`helm repo update`

and all was good...

PR #48, opened 2021-01-22
Shifted some chart links into artifacthub annotations

itzg:
I noticed [here](https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations), that the links shown on the artifacthub page can be annotated with a name.

Right now they all look like "Source" links

image

...and some are actually old/wrong.


billimek (approved):
This is great, enhancing the artifacthub experience!

PR #52, opened 2021-02-07
fix helm command

Biswajee:
Add `repo` in `helm search` command.


itzg (approved):
Thanks! It had been fixed in master, but I forget the CI build isn't sync'ing non-chart files to gh-pages.

PR #50, opened 2021-02-05
feat: add chart for bungeecord

ChipWolf:
I've attempted to follow the standards of the other charts in this repo as closely as possible, I'm sure this needs to be iterated on

Let me know if anything should be changed, this is open to edits

PR #55, opened 2021-02-21
fix(minecraft-proxy)!: use generic variable names

ChipWolf:
Replaces explicit references to bungee with proxy where necessary.

As per https://github.com/itzg/minecraft-server-charts/pull/50#issuecomment-781757398, added to the readme.

Major version bump as it's a breaking change, now using `major.minor.patch` for consistency.


ChipWolf:
Should this be a major version bump as it's a breaking change?

README.md
```diff
@@ -16,9 +16,10 @@ helm repo add itzg https://itzg.github.io/minecraft-server-charts/
 You can then run `helm search repo itzg` to see the charts.
 
 ## Charts
-tt
-* [minecraft](https://github.com/itzg/minecraft-server-charts/tree/master/charts/minecraft)
-* [minecraft-bedrock](https://github.com/itzg/minecraft-server-charts/tree/master/charts/minecraft-bedrock)
+
+* [minecraft](./charts/minecraft)
```
billimek, line 20: For these, can we use absolute URLs instead of relative URLs?  The reason for this is when this readme is presented in something like artifacthub, the links will still work.

ChipWolf: Will do

itzg (approved):
Thanks!

PR #58, opened 2021-02-25
Pvc annotations, fixed extraPort service

ishioni:
Adds PVC annotations

Edit: I just caught that the new extraPort service violates RFC 1123, as it must always be lowercase. Fixed it as well

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.0.1
+version: 3.0.2
 appVersion: SeeValues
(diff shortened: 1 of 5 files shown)


billimek (approved):

ChipWolf:
> 
> 
> Adds PVC annotations
> 
> Edit: I just caught that the new extraPort service violates RFC 1123, as it must always be lowercase. Fixed it as well

oops, thanks. I need a better linter

ishioni:
> oops, thanks. I need a better linter

Actually it's still broken, but I was fixing it lazily through the web editor without testing. Will make a PR soonish

PR #65, opened 2021-04-05
minecraft-proxy: Correct rendering with toYaml in ConfigMap

KingKeithC:
This PR corrects an issue caused by a duplicate block scaler token in the ConfigMap created when the `minecraftProxy.config` value is supplied.

The current ConfigMap template defines the `config.yml` file as follows:
```
data:
  config.yml: |
{{ toYaml .Values.minecraftProxy.config | indent 4 }}
```

Using the following `custom-values.yml` as an example:
```
minecraftProxy:
  config: |
    ---
    online_mode: true
    servers:
      lobby:
        motd: 'Testing...'
        address: minecraft-minecraft:25565
        restricted: false
```

and after rendering the template, the resulting `config.yml` file is not valid yaml:
```
keith.king at jump in ~/.../charts/minecraft-proxy on (HEAD detached at minecraft-proxy-2.0.1)*
$ helm template -f values.yaml -f custom-values.yml bungeecord .
<truncated for brevity>
---
# Source: minecraft-proxy/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: bungeecord-minecraft-proxy-config
  labels:
    app: bungeecord-minecraft-proxy
    chart: "minecraft-proxy-2.0.1"
    release: "bungeecord"
    heritage: "Helm"
data:
  config.yml: |
    |
      ---
      online_mode: true
      servers:
        lobby:
          motd: 'Testing...'
          address: minecraft-minecraft:25565
          restricted: false
<truncated for brevity>
```
Note the extra block scaler appearing on the newline, this results in `config.yml` being rendered containing the block scaler, and subsequently bungeecord fails to startup since it cannot successfully parse the config file.

After the change introduced in this PR, the template is rendered to valid yaml:
```
keith.king at jump in ~/.../charts/minecraft-proxy on fix-config*
$ helm template -f values.yaml -f custom-values.yml bungeecord .
<truncated for brevity>
---
# Source: minecraft-proxy/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: bungeecord-minecraft-proxy-config
  labels:
    app: bungeecord-minecraft-proxy
    chart: "minecraft-proxy-2.0.1"
    release: "bungeecord"
    heritage: "Helm"
data:
  config.yml:   |
    ---
    online_mode: true
    servers:
      lobby:
        motd: 'Testing...'
        address: minecraft-minecraft:25565
        restricted: false
<truncated for brevity>
```

### DIFF
--- a/charts/minecraft-proxy/templates/configmap.yaml
+++ b/charts/minecraft-proxy/templates/configmap.yaml
@@ -9,5 +9,5 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 data:
-  config.yml: {{ toYaml .Values.minecraftProxy.config | indent 4 }}
+  config.yml: {{ toYaml .Values.minecraftProxy.config | indent 2 }}
 {{- end }}
--- a/charts/minecraft-proxy/templates/configmap.yaml
+++ b/charts/minecraft-proxy/templates/configmap.yaml
@@ -9,6 +9,5 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 data:
-  config.yml: |
-{{ toYaml .Values.minecraftProxy.config | indent 4 }}
+  config.yml: {{ toYaml .Values.minecraftProxy.config | indent 4 }}
 {{- end }}
--- a/charts/minecraft-proxy/Chart.yaml
+++ b/charts/minecraft-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 2.0.1
+version: 2.0.2
 appVersion: SeeValues
 description: Minecraft proxy server (BungeeCord, Waterfall, Velocity, etc.)
 keywords:


charts/minecraft-proxy/templates/configmap.yaml
```diff
@@ -9,6 +9,5 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 data:
-  config.yml: |
-{{ toYaml .Values.minecraftProxy.config | indent 4 }}
+  config.yml: {{ toYaml .Values.minecraftProxy.config | indent 2 }}
```
itzg, line 12: Are you totally sure about the indent of 2 and not 4?

itzg (commented):
Thanks for fixing this.

charts/minecraft-proxy/templates/configmap.yaml
KingKeithC: Yes, the indent of 2 seems to produce a block that is sufficiently indented so as not to have leading spaces in the resulting `config.yml` file. It doesn't really make a difference though, the parser which reads the config file will ignore those space anyways. So it's mainly an aesthetic issue.

With `indent 4`, the resulting file doesn't conform to the indentation of the rest of the template (2 spaces):
```
keith.king at jump in ~/.../charts/minecraft-proxy on fix-config*
$ helm template -f values.yaml -f custom-values.yml bungeecord .
---
# Source: minecraft-proxy/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: bungeecord-minecraft-proxy-config
  labels:
    app: bungeecord-minecraft-proxy
    chart: "minecraft-proxy-2.0.2"
    release: "bungeecord"
    heritage: "Helm"
data:
  config.yml:     |
      ---
      online_mode: true
      servers:
        lobby:
          motd: 'Testing...'
          address: minecraft-minecraft:25565
          restricted: false
```

itzg:
@ChipWolf can you review this also? I'm sure you would have tested the use of `minecraftProxy.config` and maybe there's something missing from the values comment.

charts/minecraft-proxy/templates/configmap.yaml
itzg: ...specifically @ChipWolf , the values says "works with yaml currently", yet being shown in a block scalar

https://github.com/itzg/minecraft-server-charts/blob/4d1b55457482b4b90bcfdaf72eaf78f489ea5657/charts/minecraft-proxy/values.yaml#L104-L105

indicates that it's actually handled as "plain text". As such, I'm wondering if the intent was just an `indent 4` to transplant the user provided text content as-is, but indented 4 spaces for overall yaml correctness.

ChipWolf: > Are you totally sure about the indent of 2 and not 4?

Frankly, I'm not sure either, and I'm sure it's purely cosmetic but the [Helm docs](https://helm.sh/docs/chart_template_guide/accessing_files/) show an indentation of 6 with multiline ConfigMaps.

> @ChipWolf can you review this also? I'm sure you would have tested the use of `minecraftProxy.config` and maybe there's something missing from the values comment.

The block scalar likely landed in `values.yaml` while writing the doc after testing, so that's my bad. 😬 

> ...specifically @ChipWolf , the values says "works with yaml currently", yet being shown in a block scalar

Rethinking this, the change proposed by @KingKeithC should work for yaml & plain text.
TOML is required by Velocity, so we should probably add an additional example in `values.yaml`.

```yaml
configFilePath: /server/velocity.toml
config: |-
  config-version = "1.0"
  bind = "0.0.0.0:25577"
  motd = "Look PR 65 - it works!"
  show-max-players = 500
  ...
```

itzg (approved):
Thanks again for fixing this and glad to see it works great for the known use cases.

KingKeithC:
No problem, glad to see this was merged. Thanks!

PR #67, opened 2021-04-16
Support usage of an existing secret for rcon

RouNNdeL:
Inspired by the [grafana chart](https://github.com/grafana/helm-charts/blob/main/charts/grafana).
The ability to use an existing secret for the rcon password allows users to add the `values.yaml` to version control without exposing the password.


itzg (approved):
Thanks!

PR #68, opened 2021-04-22
Allow changing the service port

RouNNdeL:
Changing the external service port is useful when using the type `LoadBalancer` with `k3s` (maybe cloud clusters as well). 
Previously it was not possible to run multiple instances with type `LoadBalancer`, as the it does not expose the `nodePort`, but the `port` of a service.

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -162,6 +162,8 @@ minecraftServer:
   serviceType: ClusterIP
   ## Set the port used if the serviceType is NodePort
   # nodePort:
+  ## Set the external port of the service, usefull when using the LoadBalancer service type
+  # servicePort:
   loadBalancerIP:
   # loadBalancerSourceRanges: []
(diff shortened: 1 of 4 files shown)


charts/minecraft/values.yaml
```diff
@@ -162,6 +162,8 @@ minecraftServer:
   serviceType: ClusterIP
   ## Set the port used if the serviceType is NodePort
   # nodePort:
+  ## Set the external port of the service, usefull when using the LoadBalancer service type
+  # servicePort:
```
itzg, line 166: Even though there's also the default on the variable usage, I'm wondering pre-declaring the default here also would help self-document it.

itzg (commented):
Great addition. I can either way on my comment, so just want to get your thoughts.

charts/minecraft/values.yaml
RouNNdeL: Agreed, since it also makes a difference with the default service type: `ClusterIP`. I'll change it.

itzg (approved):

PR #69, opened 2021-05-02
Add support for CURSEFORGE + CF_SERVER_MOD

rtrox:
Simple PR to add support for the CURSEFORGE type via CF_SERVER_MOD.

In the interim, can work around this by using extraEnv to set CF_SERVER_MOD directly.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.1.5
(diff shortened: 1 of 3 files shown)


itzg (changes requested):
You'll just need to bump the chart's patch version and this is good to go. Thanks!

rtrox:
whoops, thanks @itzg! bumped 3.1.5 -> 3.1.6.

itzg (approved):

PR #72, opened 2021-05-05
(hotfix): extra volumes

shokohsc:
Typo in templates stopped this 'feature' to be used.


itzg:
I'm not sure how the existing structure was a problem. I just tested with:

```yaml
minecraftServer:
  eula: "TRUE"
extraVolumes:
  - volumeMounts:
      - name: worlds
        mountPath: /worlds
        readOnly: true
    volumes:
      - name: worlds
        emptyDir: {}
```

and it rendered the latter part of container and volumes section as I'd expect:
```yaml
        volumeMounts:
        - name: datadir
          mountPath: /data
        - mountPath: /worlds
          name: worlds
          readOnly: true
      volumes:
      - name: datadir
        emptyDir: {}
      - emptyDir: {}
        name: worlds
```

shokohsc:
Hello, that is... surprising. I tried those options without success with a slight difference though:

I did not specified a volumeMount for the main (minecraft) container.
Only an external volume that is used explicitly for an initContainer and I did have a syntax error about the key volumeMounts but about the main container, not the initContainer.

May I ask you to confirm that use case on your side and tell me if it fails too ?

````
extraVolumes:
  volumes:
    - name: test
      emptyDir: {}
initContainers:
  - name: test
    image: busybox
    command:
      - touch
    args:
      - /test/hello
    volumeMounts:
      - name: test
        mountPath: /test
````

the `initContainers` key is not necessary.

even specifying `extraVolumes.volumeMounts: []` did fail.
I'm suspecting my template key `range` instead of `if` to trigger the error I'm encountering

**Comment by itzg** [comment]
`extraVolumes` is a list, so you seem to be missing a "dash" in your structure:

Your snippet
```
extraVolumes:
  volumes:
    - name: test
      emptyDir: {}
```

should be
```
extraVolumes:
  - volumes:
      - name: test
        emptyDir: {}
```

PR #75, opened 2021-05-14
add nodePort to rcon service

gilesknap:
This gives the rcon service definition the same options as the minecraft service.

(I need external rcon because I'm experimenting with world building using a python rcon client https://github.com/conqp/mcwb)

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.1.6
+version: 3.1.7
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.1.7
+version: 3.1.8
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -180,6 +180,8 @@ minecraftServer:
     existingSecret:
     secretKey: rcon-password
     serviceType: LoadBalancer
+    ## Set the external port if the rcon serviceType is NodePort
+    # nodePort: 
     loadBalancerIP:
     # loadBalancerSourceRanges: []
     ## Set the externalTrafficPolicy in the Service to either Cluster or Local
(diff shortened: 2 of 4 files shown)


charts/minecraft/values.yaml
```diff
@@ -180,6 +180,8 @@ minecraftServer:
     existingSecret:
     secretKey: rcon-password
     serviceType: LoadBalancer
```
itzg, line 182: I hadn't noticed before the default is LoadBalancer. I think I'll change that to ClusterIP after merging your PR.

charts/minecraft/values.yaml
```diff
@@ -180,6 +180,8 @@ minecraftServer:
...
+    ## Set the external port if the rcon serviceType is NodePort
+    # nodePort:
```
itzg, line 184: Could actually leave that uncommented so it will be "visible in the schema"

itzg (commented):
Thanks! Your changes look great and will merge either way regarding your thoughts on the second comment.

charts/minecraft/values.yaml
gilesknap: yes this makes sense

charts/minecraft/values.yaml
gilesknap: I agree with this but I was copying how its handled in the minecraft service. Should I change that too?

itzg: Good point. Yes, please that one too.

charts/minecraft/values.yaml
billimek: Agreed, `ClusterIP` is a safer default to use IMO.  Not everyone is going to have a cluster that can handle `LoadBalancer`.

gilesknap: I've switched to ClusterIP default for rcon

charts/minecraft/values.yaml
gilesknap: I've un-commented both nodePort in values.

charts/minecraft/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.1.7
+version: 3.1.10
```
itzg, line 3: I think you'll only need to bump the version once for the overall PR

gilesknap: doh!

itzg:
Thanks for all of the additional cleanup. Un-bump the chart down to 3.1.8 and I'll get this merged.

itzg (approved):

PR #76, opened 2021-05-25
fix(minecraft): expose protocol config for extra ports

ChipWolf:
We've got a service for voice chat in-game which requires UDP, I realized this option was not exposed, so here it is.

### DIFF
--- a/charts/minecraft/templates/deployment.yaml
+++ b/charts/minecraft/templates/deployment.yaml
@@ -218,7 +218,7 @@ spec:
         {{- if .service.enabled }}
         - name: {{ .name }}
           containerPort: {{ .containerPort }}
-          protocol: TCP
+          protocol: {{ .protocol }}
         {{- end }}
         {{- end }}
         volumeMounts:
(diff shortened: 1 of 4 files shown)


charts/minecraft/templates/deployment.yaml
```diff
@@ -218,7 +218,7 @@ spec:
         {{- if .service.enabled }}
         - name: {{ .name }}
           containerPort: {{ .containerPort }}
-          protocol: TCP
+          protocol: {{ .protocol }}
```
itzg, line 221: Seems like it should default to TCP or better yet wrap `protocol` line in a conditional since the kubernetes API default is TCP when not specified.

charts/minecraft/templates/extraports-svc.yaml
```diff
@@ -35,7 +35,7 @@ spec:
   - name: {{ .name }}
     port: {{ .service.port }}
     targetPort: {{ .name }}
-    protocol: TCP
+    protocol: {{ .protocol }}
```
itzg, line 38: Same approach would be used here.

itzg (commented):
Great idea to add this. Just today I came across Simple Voice Chat's need for an additional UDP port publishing.

itzg (approved):

PR #78, opened 2021-06-07
feat(minecraft-proxy): add embedded service option to extraports

ChipWolf:
This allows you to embed the service port in the proxy service in the event you would like your extra ports to share the same service.
This is useful for sharing a single load balancer IP with MetalLB & still being able to use the Local external traffic policy to retain the origin IPs.


itzg (approved):
Thanks!

PR #81, opened 2021-07-14
Add spawn protection in helm chart value

inductor:
Hello! This PR adds spawn-protection value to be editable on Helm chart.

The original feature is based on https://github.com/itzg/docker-minecraft-server/#set-spawn-protection

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -127,6 +127,8 @@ minecraftServer:
   spawnMonsters: true
   # Determines if villagers will be spawned.
   spawnNPCs: true
+  # Sets the area that non-ops can not edit (0 to disable)
+  spawnProtection:  16
   # Max view distance (in chunks).
   viewDistance: 10
   # Define this if you want a specific map generation seed.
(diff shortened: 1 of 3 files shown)


itzg (approved):
Thanks

charts/minecraft/values.yaml
```diff
@@ -127,6 +127,8 @@ minecraftServer:
   spawnMonsters: true
   # Determines if villagers will be spawned.
   spawnNPCs: true
+  # Sets the area that non-ops can not edit (0 to disable)
+  spawnProtection:  16
```
itzg, line 131: I think the lint check is complaining about the extra space before 16

itzg (changes requested):
Lint fix needed

charts/minecraft/values.yaml
inductor: fixed. thanks!

itzg:
Sorry, there was another PR ahead of yours and you'll need a small merge and resolve...and bump the version one more time.

inductor:
I think it's a better idea to change the release number you update the version tag not on master...

itzg:
> I think it's a better idea to change the release number you update the version tag not on master...

The current strategy is annoying in rare cases like what you experienced, but seems to be standard practice. 

@billimek any thoughts to add?

inductor:
@itzg Hi. I updated the version. Can you merge it?

itzg (approved):

PR #60, opened 2021-02-26
set ingress apiversion in accordance to k8s version

ishioni:
Let's not discriminate those with older k8s versions :)
I'll rebase this once #59 hits master

PR #82, opened 2021-07-16
Fix kubeversion check in helper template

billimek:
Fixes issue introduced with #60.  It looks like `.Capabilities.KubeVersion.GitVersion` [has been deprecated](https://github.com/helm/helm/blob/2f4ef705a9e8ea0e38a2c99c29c5e4de59b03d75/pkg/chartutil/capabilities.go#L60) and we should use  `.Capabilities.KubeVersion.Version` instead.

### DIFF
--- a/charts/minecraft-proxy/Chart.yaml
+++ b/charts/minecraft-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 2.2.0
+version: 2.2.1
 appVersion: SeeValues
 description: Minecraft proxy server (BungeeCord, Waterfall, Velocity, etc.)
 keywords:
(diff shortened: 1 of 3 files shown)


ishioni:
@billimek Right, I knew this would be needed at some point, but very helpfully helm doesn't preserve it's old pre-stable docs except for 2.7, so I can't go back and see when they made the switch/deprecated the old way

itzg (approved):

billimek:
This is still failing,

```
error calling include: template: minecraft/templates/_helpers.tpl:19:44: executing "minecraft.ingress.apiVersion" at <.Capabilities.KubeVersion.Version>: nil pointer evaluating interface {}.KubeVersion
```

Checking for what can be a proper fix.

ishioni:
@billimek Ah i didn't notice, this is now a range! You need to call it with $, like the other ones

PR #83, opened 2021-07-16
Fix syntax error with ingress changes

billimek:
This fixes #82.

Thanks to @ishioni for the catch that the ingress object is wrapped in a range and the syntax of the include needs to be a bit different!


itzg (approved):

ChipWolf:
will need to port this to minecraft-proxy

PR #86, opened 2021-10-28
Adding support for subPath

gandazgul:


gandazgul:
@itzg thanks, sorry it took me so long to come back to this. I bumped the chart version.

charts/minecraft/templates/deployment.yaml
```diff
@@ -239,6 +239,9 @@ spec:
         volumeMounts:
         - name: datadir
           mountPath: /data
+          {{- if (and .Values.persistence.dataDir.enabled .Values.persistence.dataDir.subPath) }}
+          subPath: {{ .Values.persistence.dataDir.subPath }}
```
billimek, line 243: Should we also add a placeholder in the `values.yaml` file with an example (commented) of what this setting would be.

Generally speaking, if we reference a value, it's a good practice to have an example in the values.yaml file.

itzg: Good point. Can you add that @gandazgul ?

gandazgul: This is done

itzg (approved):

carlos-ravelo-ask:
thanks!

PR #87, opened 2021-11-12
Include `envFrom` and `env.value.valueFrom` in deployments

### LINKED ISSUE #71: [minecraft] allow valueFrom for extraEnv
I have a couple of use cases that require me to inject extra environment variables which require the use of `valueFrom`.

This could be implemented into the chart by exposing an additional variable that accepts the full deployment `spec.template.spec.containers[].env` syntax. This would only require a minor release.

```yaml
someNewExtraEnvVariable:
  - name: NORMAL_VAR
    value: "foobar"
  - name: HOST_IP
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
```

Alternatively, the less messy but slightly less user-friendly option would be to allow the full use of the syntax with the `extraEnv` variable, this would likely be a breaking change unless someone can think up the logic to allow this kind of syntax:

```yaml
extraEnv:
  USE_AIKARS_FLAGS: "true"
  HOST_IP:
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
 ```
 
Thoughts?

albertrdixon:
## What

- Add optional `envFrom` to deployments
- Update `extraEnv` to allow for setting `valueFrom` in env var specs

Fixes #71

### DIFF
--- a/charts/minecraft-bedrock/Chart.yaml
+++ b/charts/minecraft-bedrock/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-bedrock
-version: 1.1.9
+version: 1.2.0
 appVersion: 1.16
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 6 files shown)


itzg (approved):
Thanks!

PR #88, opened 2021-11-14
Add ingressClassName to templates

Terag:
Add the field `ingressClassName` into the Ingress oject generated from the Chart templates. This is to support the use of IngressClass objects that replace the use of annotations (https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class) when specifying the Ingress Controller to use.

The behavior can be tested using the helm chart from (https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx) `v4.0.6` with this example `values.yaml`:
```yaml
controller:
  watchIngressWithoutClass: false
  ingressClassByName: true
  electionID: ingress-controller-custom
  ingressClassResource:
    name: custom-nginx
    enabled: true
    default: false
    controllerValue: "k8s.io/custom-ingress-nginx"
  service:
    externalIPs: 
      - *your_ip*
  replicaCount: 1
```

### DIFF
--- a/charts/minecraft-proxy/Chart.yaml
+++ b/charts/minecraft-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 2.2.0
+version: 2.3.0
 appVersion: SeeValues
 description: Minecraft proxy server (BungeeCord, Waterfall, Velocity, etc.)
 keywords:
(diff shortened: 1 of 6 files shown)


itzg (changes requested):
Changes look good. Bump the minor version in the charts and this will be good to go.

Terag:
Minor versions bumped.

itzg (approved):

PR #89, opened 2021-12-02
Fixed small typo in local port forwarding

jbarratt:
The echo command returned the wrong port (22565 instead of 25565)


itzg (changes requested):
Wow, that's been wrong for quite a while. Thanks for fixing that.

Bump the patch version on the chart and this is good to merge:

https://github.com/itzg/minecraft-server-charts/runs/4390251752?check_suite_focus=true#step:7:23

jbarratt:
Ok, done! No problem. It just jumped out at me after I did the first deploy.

PR #90, opened 2021-12-08
Addded a way to add a service account to the minecraft-proxy allowing…

ArturoGuerra:
… for easier access to the kubernetes api


itzg (approved):
Thanks

PR #97, opened 2022-01-19
 Allow adding sidecarContainers

Robbilie:
https://github.com/itzg/minecraft-server-charts/pull/96

Changed the branch in my repo and that didnt go as planned…

PR #98, opened 2022-01-20
Allow setting deployment replica count

Robbilie:
In by its own there will probably be not that many people that have a usecase for this, i dont think it hurts to have it configurable though, keeps me from making a hard fork of this chart for now :P

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -3,6 +3,9 @@ image: itzg/minecraft-server
 imageTag: latest
 imagePullSecret: ""
 
+# ### WARNING ###
+# Minecraft is not horizontally scalable, adjusting this will most likely break your setup.
+# ### WARNING ###
 replicaCount: 1
 
 ## Configure resource requests and limits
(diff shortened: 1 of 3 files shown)


charts/minecraft/values.yaml
```diff
@@ -3,6 +3,8 @@ image: itzg/minecraft-server
 imageTag: latest
 imagePullSecret: ""
 
+replicaCount: 1
```
itzg, line 6: It feels like this needs a big warning similar to what you described at the top of the PR.

Robbilie: you think a comment in the values.yaml is enough? ill add one

Robbilie: there we go

itzg: Thanks!

itzg (approved):
Enjoy!

Robbilie:
i got one more thing coming :D

Robbilie:
https://github.com/itzg/minecraft-server-charts/pull/99

ChipWolf:
As a point of interest @Robbilie, why did you add this?

Robbilie:
@ChipWolf I mentioned it in the other PR

https://github.com/itzg/minecraft-server-charts/pull/97

I wanna run a synchronization sidecar, that's why I added some features to the chart, got another one coming at some point for podlabels

PR #103, opened 2022-01-27
Update RClone Config

### LINKED ISSUE #100: rclone config should be provided as a Secret
An rclone config file contains data that should be kept secret, such as API keys for storage backends. Currently, the supported way to set the rclone config file contents is to put these contents into the values for a release. The contents will then be stored in a ConfigMap, which is not a secure place to put secrets.

It's currently possible to override this behaviour with a safer approach, if you set up the volume mounts using `extraVolumes` and create the secret manually. Here's what I'm using:
```
    extraVolumes:
      - volumeMounts:
        - name: real-rclone-config
          mountPath: /config/rclone
        volumes
        - name: real-rclone-config
          secret:
            secretName: my-rclone-config-secret
```

However, it's probably not a good idea for the chart/documentation to be encouraging users to use an unsafe configuration. The ConfigMap-based approach should be deprecated or transparently replaced with a Secret-based approach. This chart should also allow users to specify an existing secret, similarly to how the RCON password is handled; this is especially useful for GitOps-style workflows (e.g. Flux) where secrets have to be created separately from Helm release inputs.

merkinbob:
Switch rclone.config `configmap` to a secret, also add `rcloneConfigExistingSecret` as an option as requested.

fixes #100

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.6.6
+version: 3.7.2
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/templates/deployment.yaml
+++ b/charts/minecraft/templates/deployment.yaml
@@ -358,8 +358,12 @@ spec:
       {{- end }}
       {{- if eq .Values.mcbackup.backupMethod "rclone" }}
       - name: rclone-config
-        configMap:
-          name: {{ template "minecraft.fullname" . }}-rclone-config
+        secret:
+        {{- if .Values.mcbackup.rcloneConfigExistingSecret }}
+          secretName: {{ .Values.mcbackup.rcloneConfigExistingSecret }}
+        {{- else }}
+          secretName: {{ template "minecraft.fullname" . }}-rclone-config
+        {{- end }}
           items:
           - key: rclone.conf
             path: rclone.conf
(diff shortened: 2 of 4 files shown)


merkinbob:
@wlritchi haven't tested this yet, still a WIP

wlritchi:
:+1: This should meet my use case, so I'm happy. Minor note: it looks like even if `rcloneConfigExistingSecret` is provided, the WIP will still create a Secret (presumably with nothing in it) - maybe add a conditional there?

Also, is there any automation in this repo that builds+tags charts from PRs so I can point my Helm at it to test? If not, I can do it manually, but I figured I should ask in case there's an easier way already set up.

merkinbob:
@wlritchi great catch, updated.  My fork is hosting the chart for my testing needs.  Feel free to test it out.

https://github.com/merkinbob/minecraft-server-charts/releases/tag/minecraft-3.7.2

wlritchi:
I was hoping to just point Helm at the GitHub Pages URL. It looks like the releaser action is running and updating the `gh-pages` branch, but that's not reflected in https://merkinbob.github.io/minecraft-server-charts/index.yaml. Is your repo set up to use some other branch for GitHub Pages?

merkinbob:
i see 3.7.2 on https://merkinbob.github.io/minecraft-server-charts/index.yaml

merkinbob:
@wlritchi turned it off and on again, it's there now

wlritchi:
I tested it and it seems to be working :) thanks!

itzg (approved):

itzg:
Thanks both of you for fixing/testing. 

@merkinbob set this to non-draft when you're ready

PR #105, opened 2022-02-06
Add rcloneDestDir param

### LINKED ISSUE #104: rclone backup writes to root directory instead of mounted emptyDir
Context
-----
This chart creates an emptyDir (or, if backup persistence is enabled, a PVC) mount at the location given by `destDir`. Kubernetes sets the filesystem permissions automatically so that `tar` backups can be written to this directory.

The `destDir` option is doing double duty; for `rclone` backups, it also specifies the location *within the rclone bucket* where backups will be uploaded.

Issue
-----
When `rclone` is selected as the backup method, the backup tar file is temporarily written to disk before being uploaded. However, the file is written in the current working directory, which defaults to `/`. On some (all?) Kubernetes providers, the root directory is not writable to containers that have dropped privileges, so unless the container has been set to run as root, the temporary file cannot be written, and the backup fails.

There's a few options for fixes:

1. Either this chart or the mc-backup Dockerfile should be updated to set the working directory to `destDir`; or
2. The script should be updated to write `rclone` temporary files to `destDir`; and/or
3. A new parameter should be added to disambiguate "backup working directory" from "rclone destination path".

It feels cleanest to decouple the two uses of `destDir`, but doing so could also be a breaking change for people who have `rclone` backup already working via a root-privileged container.

wlritchi:
Companion PR to itzg/docker-mc-backup#63. Add an `rcloneDestDir` param to the chart, mapped to the `RCLONE_DEST_DIR` environment variable for `mc-backup`.

Fixes #104.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 3.7.2
+version: 3.8.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/templates/deployment.yaml
+++ b/charts/minecraft/templates/deployment.yaml
@@ -92,6 +92,8 @@ spec:
         {{- if eq .Values.mcbackup.backupMethod "rclone" }}
         - name: RCLONE_REMOTE
           value: {{ .Values.mcbackup.rcloneRemote | quote }}
+        - name: RCLONE_DEST_DIR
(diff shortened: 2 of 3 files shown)


wlritchi:
Is it correct to bump the chart version in a PR like this? Not quite sure what the workflow is supposed to be for the helm chart releaser.

itzg:
> Is it correct to bump the chart version in a PR like this? Not quite sure what the workflow is supposed to be for the helm chart releaser.

For now, yes it's actually required that PRs bump the version.

itzg (approved):
Great pair of PRs. Thanks!

PR #110, opened 2022-03-06
fix: Check if extraVolumes.volumeMounts is defined

eaglesemanation:
It could be useful to create volumes without actually mounting them in
case if they are used exclusively for init containers


itzg (commented):
Changes look good. The linter is wanting the versions to be bumped. Patch version seems fine.

Issue #112, opened 2022-03-14
extraVolumes: volumeMounts causes null in output for deployment volumes.


tomp736:
I am trying to add extraVolumes to attach an nfs share. When I have volumes and volumeMounts in extraVolumes, I am getting a null string in the output. I did a few tests and it looks like the issue is the helm chart - specifically the area that is adding volumes. When looping through extraVolumes, each volumeMount element will show as null.

This should be a simple fix - only loop and add volumes from extraVolumes.

## With volumeMounts in extra volumes.

***values.yaml***
``` values.yaml
minecraftServer:
  serviceType: LoadBalancer
  eula: true
  type: "CURSEFORGE"
  cfServerMod: "/modpacks/All+the+Mods+6+-+To+the+Sky+-1.2.0.zip"

persistence:
  storageClass: cstor-csi-disk
  dataDir:
    enabled: true
    Size: 5Gi

extraVolumes:
  - volumeMounts:
      - name: modpacks
        mountPath: "/modpacks"
        readOnly: true
  - volumes:
      - name: modpacks
        nfs:
          server: nfs-server
          path: "/nfs/modpacks"
```

Running helm template --debug

``` bash
helm template minecraft --values minecraft_values.yaml minecraft-server-charts/minecraft --debug > template.out
install.go:178: [debug] Original chart version: ""
install.go:195: [debug] CHART PATH: /root/.cache/helm/repository/minecraft-3.8.2.tgz

Error: YAML parse error on minecraft/templates/deployment.yaml: error converting YAML to JSON: yaml: line 145: could not find expected ':'
helm.go:84: [debug] error converting YAML to JSON: yaml: line 145: could not find expected ':'
YAML parse error on minecraft/templates/deployment.yaml
``` 

***out.yaml***
``` out.yaml
       volumeMounts:
        - name: datadir
          mountPath: /data
        - name: backupdir
          mountPath: /backups
          readOnly: true
        - mountPath: /modpacks
          name: modpacks
          readOnly: true
      volumes:
      - name: datadir
        persistentVolumeClaim:
          claimName: minecraft-1-minecraft-datadir
      - name: backupdir
        emptyDir: {}
      null
      - name: modpacks
        nfs:
          path: /nfs/modpacks
          server: nfs-server
```

## After removing the volumeMounts from extraVolumes:

***values.yaml***
``` values.yaml
minecraftServer:
  serviceType: LoadBalancer
  eula: true
  type: "CURSEFORGE"
[...]

PR #113, opened 2022-03-14
extravolumes: extraVolumes - check if element is a volume before writing

tomp736:
check if is volumes before writing
#112


itzg (commented):
Change looks great. Just needs the patch version bumped.

itzg:
Merge conflict on the version needs to be fully resolved

PR #114, opened 2022-03-17
feat: SPIGET_RESOURCES env var

eaglesemanation:


charts/minecraft/templates/deployment.yaml
```diff
@@ -266,6 +266,10 @@ spec:
           value: "TRUE"
         {{- end }}
         {{- end }}
+        {{- if .Values.minecraftServer.spigetResources }}
+        - name: SPIGET_RESOURCES
+          value: {{ include "helm-toolkit.utils.joinListWithComma" .Values.minecraftServer.spigetResources }}
```
itzg, line 271: Please use a `join ","`

itzg (changes requested):
Thanks for adding this

charts/minecraft/templates/deployment.yaml
eaglesemanation: Then should I update [minecraft-proxy](https://github.com/itzg/minecraft-server-charts/blob/2fea96dd82706f6b083b30f5dd9848b10d8b43c6/charts/minecraft-proxy/templates/deployment.yaml#L100) as well?

charts/minecraft-proxy/templates/deployment.yaml
```diff
@@ -97,7 +97,7 @@ spec:
         {{- end }}
         {{- if .Values.plugins }}
         - name: PLUGINS
-          value: {{ include "helm-toolkit.utils.joinListWithComma" .Values.plugins }}
+          value: {{ join "," .Values.plugins }}
```
eaglesemanation, line 100: This path seems to be incorrect, because in [values.yaml](https://github.com/itzg/minecraft-server-charts/blob/2fea96dd82706f6b083b30f5dd9848b10d8b43c6/charts/minecraft-proxy/values.yaml#L85) it's written as `.Values.minecraftProxy.plugins`

charts/minecraft-proxy/templates/deployment.yaml
```diff
@@ -95,9 +95,9 @@ spec:
           value: {{ .Values.minecraftProxy.jarRevision }}
         {{- end }}
         {{- end }}
-        {{- if .Values.plugins }}
+        {{- if .Values.minecraftProxy.plugins }}
         - name: PLUGINS
-          value: {{ include "helm-toolkit.utils.joinListWithComma" .Values.plugins }}
+          value: {{ join "," .Values.minecraftProxy.plugins }}
```
itzg, line 100: Thanks for taking care of this one too.

itzg (approved):
Thanks!

Issue #115, opened 2022-03-17
Feature Request: Add chart for rcon-web-admin


eaglesemanation:
I would like to write a chart for your rcon-web-admin container, just wanted to ask if it's ok if I'll contribute it here instead of doing it separately.

itzg:
Yes that seems appropriate to contribute here. Thanks!

PR #117, opened 2022-03-21
Rcon Web

eaglesemanation:
Chart for `itzg/rcon` deployment, implements #115

### DIFF
--- a/charts/rcon-web-admin/README.md
+++ b/charts/rcon-web-admin/README.md
@@ -1,0 +1,61 @@
+# RCON Web Admin
+
+RCON Web Admin is a powerful web interface to control your RCON servers. All servers with RCON support should work.
+
+This application can install and run on a server, Raspberry Pi, or other devices which are online 24/7.
+
+Out of the box, RCON Web Admin can check users for high ping, VAC status, or filter the chat for you around the clock.
+
+## Introduction
+
+This chart creates a single RCON Web Admin Pod, plus Services for HTTP and WebSocket traffic.
+
+## Prerequisites
+
+- 512 MB of RAM
+- Kubernetes 1.4+ with Beta APIs enabled
+- Ingress Controller or Load Balancer available on Kubernetes cluster
+
+## Installing the Chart
+
+To install the chart with the release name `rcon-web` and expose it through load balancer, run:
+
+```shell
+helm install \
+  --set rconWeb.password=admin_password,service.type=LoadBalancer \
+  rcon-web itzg/rcon-web-admin
+```
+
+This command deploys a RCON Web Admin dedicated server with sensible defaults.
+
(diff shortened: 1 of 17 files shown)


itzg (commented):
This looks great! Feel free to mark the PR ready when you are.

itzg:
Looks like the test setup needs to be tweaked

https://github.com/itzg/minecraft-server-charts/runs/5653804720?check_suite_focus=true#step:9:192

charts/rcon-web-admin/templates/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+              value: {{ .Values.rconWeb.websocketRcon | default false | ternary "TRUE" "FALSE" | quote }}
+            {{- if .Values.ingress.enabled }}
+            {{ $wsUrl := print .Values.ingress.host (trimSuffix "/" .Values.ingress.path) "/websocket" }}
+            - name: RWA_WEBSOCKET_URL_SSL
+              value: {{ print "wss://" $wsUrl | quote }}
+            - name: RWA_WEBSOCKET_URL
+              value: {{ print "ws://" $wsUrl | quote }}
+            {{- end }}
+          {{- if not .Values.ingress.enabled }}
+          command:
```
eaglesemanation, line 73: I don't really like this whole part, but there is unfortunate loop of "To deploy this chart you need to configure websocket url, to get websocket url you need to deploy this chart". Better solution would be to resolve it upstream, but for now this should work fine

itzg: Understood and agreed that I don't like how rcon-web handles websockets in the first place with the requirement of a dedicated port.

eaglesemanation:
Sorry, tried to recreate github workflows locally with `act`, but it kept breaking, so I had to spam small changes to see if they work.

itzg:
No problem at all. I do the same.

charts/rcon-web-admin/README.md
```diff
@@ -0,0 +1,61 @@
...
+Out of the box, RCON Web Admin can check users for high ping, VAC status, or filter the chat for you around the clock.
+
+## Introduction
+
+This chart creates a single RCON Web Admin Pod, plus Services for HTTP and WebSocket traffic.
+
+## Prerequisites
+
+- 512 MB of RAM
+- Kubernetes 1.4+ with Beta APIs enabled
```
eaglesemanation, line 16: I only tested with v1.23.4+k3s1 and kind v1.23.1, so I have no clue which versions are supported :)

itzg: Yeah 1.4 seems ancient. I don't mind leaving it or going with 1.23 since that's confirmed.

itzg (approved):
All looks great!

PR #121, opened 2022-05-05
Update charts to standard image.repo and image.tag

brandon099:
This PR updates the charts to utilize the current standard format for the image definition:

```yaml
image:
  repository:
  tag:
  pullPolicy:
```

It should be non-breaking for those using the default `values.yaml` for the image portion of the chart.

What this does allow for, since it's the Helm standard format, is automation that is build around this standard for helm charts can be utilized.


billimek:
As this is indeed a 'breaking' change to existing chart users who set those values, I suggest bumping the major version to indicate as such.  What do you think @itzg?

PR #124, opened 2022-06-15
Support Resource Packs

SavePointSam:
### Description:

- Fix template error when only passing a single Spigot Resource
- Add chart support for Vanilla Tweaks
- Add chart support for server resource pack


itzg (commented):
Changes look great. Just need to bump the version, like 4.1.0

SavePointSam:
Updated!

PR #125, opened 2022-06-30
Rename `spigetResources` in `minecraft` chart to `spigotResources`

Matthew-Beckett:
This PR addresses an incorrectly named value in the `minecraft` Helm chart called `spigetResources` which pulls Spigot plugins from spigotmc.org.

It also bumps the chart version to `5.0.0` as I believe this change should be considered breaking as I suspect many users would have encountered this typo and simply worked around it by using `spigetResources`.

Existing deployments with locked chart versions will be unaffected by this change, and deployments backed by PVC will also be unaffected.

However for new plugins to be installed when transitioning to version `5.0.0` on existing deployments users will need to use the new `spigotResources` value instead of `spigetResources`.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.1.0
+version: 5.0.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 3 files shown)


itzg:
The naming is correct and nobody is confused about their own names. https://spiget.org/ is an API independent of Spigot that enables retrieval of https://www.spigotmc.org/ plugins. I'm guessing "Spiget" was a play on words "Spig-GET"

Matthew-Beckett:
> The naming is correct and nobody is confused about their own names. https://spiget.org/ is an API independent of Spigot that enables retrieval of https://www.spigotmc.org/ plugins. I'm guessing "Spiget" was a play on words "Spig-GET"

I mean, I think it's fair to call it confusing as a user to see Spigot everywhere and then encounter Spiget once in this esoteric instance...

PR #127, opened 2022-08-07
BUGFIX: Resource Pack support on minecraft-server chart

Tuxprogrammer:
Fixed a bug where `RESOURCE_PACK`, `RESOURCE_PACK_SHA1`, and `RESOURCE_PACK_ENFORCE` would not be set unless the user specified `vanillaTweaksShareCodes`.

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.1.0
+version: 4.2.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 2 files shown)


Tuxprogrammer:
@itzg can you take a look at this change and see if this is the intended functionality for your latest chart version? I'm trying to stand up a vanilla server with a mandatory modpack and having difficulty with the chart values not setting properly. Thanks!

itzg (commented):
Thanks! Agreed, those parameters shouldn't be grouped together. 

The build validation will want the chart version bumped. Seems like bumping the minor version (middle position) makes most sense.

Tuxprogrammer:
Fixed in chart.yaml, please let me know if I need to do anything else here!

PR #139, opened 2022-11-28
Fix annotations parsing

Intellium:
This should fix annotation parsing on the service, so that multiple annotations can be specified

### DIFF
--- a/charts/minecraft-bedrock/templates/minecraft-svc.yaml
+++ b/charts/minecraft-bedrock/templates/minecraft-svc.yaml
@@ -7,8 +7,12 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- if .Values.serviceAnnotations }}
   annotations:
-    {{ toYaml .Values.serviceAnnotations | indent 4 }}
+    {{- range $key, $value := .Values.serviceAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
 spec:
 {{- if (or (eq .Values.minecraftServer.serviceType "ClusterIP") (empty .Values.minecraftServer.serviceType)) }}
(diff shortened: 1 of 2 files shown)


itzg:
Can you provide an example of an error that is reported without the change?

Intellium:
sure.

`Error: YAML parse error on minecraft/charts/minecraft-bedrock/templates/minecraft-svc.yaml: error converting YAML to JSON: yaml: line 11: did not find expected key`

This is with 2 annotations set, like:

```yaml
serviceAnnotations: 
    external-dns.alpha.kubernetes.io/hostname: minecraft.example.com
    my.custom/type: private
```

It seems to render the service template with wrong indentation.
Example:

```yaml
annotations:
        external-dns.alpha.kubernetes.io/hostname: minecraft.example.com
    my.custom/type: private
```

charts/minecraft-bedrock/templates/minecraft-svc.yaml
```diff
@@ -7,8 +7,12 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- if .Values.serviceAnnotations }}
   annotations:
-    {{ toYaml .Values.serviceAnnotations | indent 4 }}
+    {{- range $key, $value := .Values.serviceAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
```
itzg, line 14: I believe the other way to fix the extra indentation was to apply chomping to the left

```
    {{- toYaml .Values.serviceAnnotations | indent 4 }}
```

...however, your change is more consistent with similar sections in the templates, so the current change is best.

itzg (commented):
Just need the chart version bumped and then this is ready for a merge.

itzg:
...and thanks for providing the additional information.

Intellium:
Bumped the version

zetneteork (commented):
LGTM

PR #140, opened 2023-01-24
Enable tty/stdin

sfxworks:
This will allow users to access and interact with their console using `kubectl attach` 

Example in our chart repo: https://github.com/mcserverhosting-net/charts/blob/master/chart/forge-deployment/templates/server-statefulset.yaml#L57-L58

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.4.0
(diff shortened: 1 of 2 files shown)


itzg (commented):
Thanks! I'm surprised it wasn't there from the start 😀

Can you bump the chart's patch version? That will make the linter happy.

sfxworks:
Np there ya go

billimek (approved):
LGTM

Issue #141, opened 2023-01-24
Missing values in helm templates


zetneteork:
In the helm chart, two values are missing, as previously added in Minecraft bedrock.
https://github.com/itzg/docker-minecraft-bedrock-server/blob/ab51be217cf5f245fcb3239466f145af3a9422ba/property-definitions.json#L112

https://github.com/itzg/docker-minecraft-bedrock-server/blob/ab51be217cf5f245fcb3239466f145af3a9422ba/property-definitions.json#L116

I did not find those two values can be easily controlled in the helm chart values. 

```yaml
  "emit-server-telemetry": {
    "env": "EMIT_SERVER_TELEMETRY",
    "allowed": ["true","false"]
  },
  "enable-lan-visibility": {
    "env": "ENABLE_LAN_VISIBILITY",
    "allowed": ["true","false"]
  }
```

I expect this feature can be added.

itzg:
You can use `extraEnv` to configure those. 

I would accept a PR to associate specific values if you or anyone else is interested.

zetneteork:
https://github.com/itzg/minecraft-server-charts/pull/142

zetneteork:
> You can use `extraEnv` to configure those.
> 
> I would accept a PR to associate specific values if you or anyone else is interested.

It would be useful to list available options suitable for bedrock `extraEnv` in README file.

itzg:
Interesting idea. A link to the properties mapping file would be a nice compromise. Want to include that with your PR?

nilsrenes:
Hi , do you have a link for adding custom env options via helm kubernetes ? Do I just add them to the Charts.yml?

PR #142, opened 2023-01-24
Missing values in helm templates #141

zetneteork:
Based on https://github.com/itzg/minecraft-server-charts/issues/141 I am pushing this feature.


itzg (commented):
Thanks. Can you bump the patch version on the chart? With that it's good to merge.

Issue #143, opened 2023-02-22
Max world size inconsistent between Docker image and Helm default values.yaml


JJGadgets:
I am switching my Minecraft from my Docker host to my Kubernetes cluster. 
On the Docker side, I can't find any defaults for the envvar of the maximum world size, and when `/worldborder get` is run in-game, it shows what I understand to be the Minecraft default of 5999968, or roughly 60 million.
Meanwhile, the Helm Chart's default values.yaml shows that the Helm default is set to `maxWorldSize: 10000`.

This could cause an imported world to be significantly altered to fit the new constraints, which is undesired.
Could the default value for `maxWorldSize` be changed to the Minecraft default or at least be consistent with the behavior observed when the same image is deployed on Docker?

itzg:
Oh wow, I didn't notice all of those were preset to non-default values. I inherited this chart so was long before my time. ALL of those values need to be made conditional, which looks like a effort 😡. 

I don't even use Helm, so I would recommend either managing the manifest files directly or using the built in kustomize basing off of https://github.com/itzg/docker-minecraft-server/tree/master/kustomize

JJGadgets:
For me personally I've set this value manually in my HelmRelease (I use FluxCD rather than `helm install` directly): https://github.com/JJGadgets/Biohazard/blob/main/kube/3-deploy/2-apps/minecraft/3-install.yaml#L35

Would you be open to a PR? I'll see what I can do, as far as I see there are others using this Helm Chart too so it would be beneficial.

itzg:
Definitely open to a PR. I'm pondering whether 

- an unset value would indicate the server.properties default should be left as is. Downside is that values.yaml discovery is lessened since all those lines would need to be commented out by default
- a special value like "default" could be specified. Downside is that it will look a little odd especially for numerical values

Either way the repetition in deployment.yaml hurts my eyes. With either of the above I'd like a template-macro to be used that is given the values variable and corresponding environment variable name. The conditional can be implemented once in the macro. 

Actually, I see you using a lot of extraEnv entries. I'd really prefer it all be done that way since the maintenance of value mappings here in the chart is highly redundant maintenance.

JJGadgets:
> Actually, I see you using a lot of extraEnv entries. I'd really prefer it all be done that way since the maintenance of value mappings here in the chart is highly redundant maintenance. 

[...]

PR #145, opened 2023-03-04
feat(minecraft): allow values to fall back to server.properties default

itzg:
This solves these things:
- simplifies the ugly, repetitive two lines of env name-values
- provides a common mechanism to allow for null'ing out a value, such as `--set minecraftServer.maxWorldSize=`, leaving out the env var, and allowing `server.properties` value to be used instead.
- gets rid of inconsistent/redundant use of `default ""` clauses

@billimek , if this looks good to you, then I'll do the same for the other three charts.

For #143

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.6.0
+version: 4.6.1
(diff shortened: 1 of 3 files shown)


billimek (approved):
LGTM, sorry for the delay in reviewing!

itzg:
That's ok, I had forgotten I put this change out there 😀

PR #146, opened 2023-04-01
Extra labels, Modrinth, CurseForge and regural URLs

eaglesemanation:
Originally I wanted to just add labels, but then I went a bit overboard and added everything that I had to define through extra env vars for my own servers. Also added a schema file that will verify if values are valid, I'll leave it to your judgment to decide if that's something that you want.


itzg (commented):
Looks great and the schema inclusion is very nice. Bump the minor version and then this will be ready to merge.

itzg (approved):
Thanks!

PR #148, opened 2023-04-02
fix: Actually join lists where needed

eaglesemanation:
Seems like my brain doesn't want to work on weekend, sorry for spam. Missed join function in few places, missed modUrls during merge conflict resolve. I've tested with `helm template .`, all new values look correct, hope this is last PR for this change =)


billimek (approved):

itzg (approved):
No problem at all 😀

PR #149, opened 2023-04-07
Fix the use of required call for FTBA packs

ananace:
Without this fix, installing an FTBA pack results in;
```
Helm install failed: template: minecraft/templates/deployment.yaml:189:54: executing "minecraft/templates/deployment.yaml" at <required>: wrong number of args for required: want 2 got 0
```


itzg (changes requested):
Thanks! Can you also bump the patch version of the chart?

itzg (approved):

PR #155, opened 2023-06-01
use strict security context

### LINKED ISSUE #154: Use strict security context
The current default values do not comply with the restricted pod security policy.

See:
- https://kubernetes.io/docs/concepts/security/pod-security-standards/.
- https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

Something like this should be fine:

```yaml
podSecurityContext:
  runAsUser: 1000
  runAsGroup: 3000
  runAsNonRoot: true
  fsGroup: 2000
  seccompProfile:
    type: RuntimeDefault

securityContext:
  capabilities:
    drop:
      - ALL
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
```

I've tested this security context and it works as expected, assuming `/tmp` is mounted as writable.

uhthomas:
Use a strict security context by default to comply with the restricted pod security policy.

Fixes #154

### DIFF
--- a/charts/minecraft-bedrock/values.yaml
+++ b/charts/minecraft-bedrock/values.yaml
@@ -22,6 +22,21 @@ tolerations: []
 
 affinity: {}
 
+podSecurityContext:
+  runAsUser: 1000
+  runAsGroup: 3000
+  runAsNonRoot: true
+  fsGroup: 2000
+  seccompProfile:
+    type: RuntimeDefault
+
+securityContext:
+  capabilities:
+    drop:
+      - ALL
+  readOnlyRootFilesystem: true
+  allowPrivilegeEscalation: false
+
 # Most of these map to environment variables. See Minecraft for details:
 # https://hub.docker.com/r/itzg/minecraft-server/
 livenessProbe:
--- a/charts/minecraft-proxy/templates/deployment.yaml
+++ b/charts/minecraft-proxy/templates/deployment.yaml
@@ -28,8 +28,7 @@ spec:
         - name: {{ .Values.image.pullSecret }}
       {{- end }}
       securityContext:
-        runAsUser: {{ .Values.securityContext.runAsUser }}
-        fsGroup: {{ .Values.securityContext.fsGroup }}
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
       initContainers: {{- toYaml .Values.initContainers | nindent 8 }}
       containers:
       - name: {{ template "proxy.fullname" . }}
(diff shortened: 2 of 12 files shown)


itzg (commented):
Do you mind reverting the auto whitespace changes?

itzg:
...and maybe related https://github.com/itzg/minecraft-server-charts/actions/runs/5146374308/jobs/9265335257?pr=155#step:7:61

itzg:
> Looks like CI fails because there is nothing mounted to the required volume `/opt/rcon-web-admin/db` as stated in https://github.com/itzg/docker-rcon-web-admin.
> 
> Is it okay for this to be an ephemeral directory, or does it require persistence?

Ephemeral is totally fine. I guess that chart hasn't been re-built in a long time.

uhthomas:
> > Looks like CI fails because there is nothing mounted to the required volume `/opt/rcon-web-admin/db` as stated in https://github.com/itzg/docker-rcon-web-admin.
> > Is it okay for this to be an ephemeral directory, or does it require persistence?
> 
> Ephemeral is totally fine. I guess that chart hasn't been re-built in a long time.

Thanks, all looks good now I think!

charts/minecraft-proxy/templates/deployment.yaml
```diff
@@ -151,6 +150,8 @@ spec:
         {{- end }}
         {{- end }}
         volumeMounts:
+        - name: tmp
```
uhthomas, line 153: I wonder if this is actually necessary for the proxy?

itzg: It seems best to include since

https://github.com/itzg/docker-bungeecord/blob/608f06bc983a3b5084e54ec299477d18089ab8e7/run-bungeecord.sh#L152

itzg:
Ready for me to merge?

uhthomas:
> Ready for me to merge?

Yep, all looks good to me.

charts/minecraft-bedrock/values.yaml
```diff
@@ -22,6 +22,21 @@ tolerations: []
 
 affinity: {}
 
+podSecurityContext:
+  runAsUser: 1000
+  runAsGroup: 3000
+  runAsNonRoot: true
+  fsGroup: 2000
```
itzg, line 29: @uhthomas I had missed this before: why did you choose group ID 3000 to run and group ID for the filesystem? It seems to be causing issues like #164

uhthomas: This is standard as recommended by the official Kubernetes documentation.

https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

`uid=1000 gid=3000 groups=2000`

The idea is that there is as little permission as possible.

itzg: Thanks for following up.

PR #158, opened 2023-06-23
feat: Add container lifecycle to the main Minecraft container

allanger:
It might be useful when a user wants to install plugins using `initContainers`, but those plugins require running some commands in order to be configured.

For example, I'm installing plugins in an `emptyDir` volume, so they are reinstalled on each pod restart. But since configs are stored in the same dir, I can't just edit them there, because they exist on that `emptyDir` volume as well. So the idea of this PR is to let users do something like:

```YAML
  spec:
    containers:
      name: minecraft-minecraft
      lifecycle:
        postStart:
          exec:
            command:
            - bash
            - -c
            - for i in {1..100}; do mc-health && break || sleep 20; done && mc-send-to-console
              say hello world
```

### DIFF
--- a/charts/minecraft/templates/deployment.yaml
+++ b/charts/minecraft/templates/deployment.yaml
@@ -142,6 +142,24 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         tty: true
         stdin: true
+        # Working here
+        {{- if or (.Values.lifecycle.postStart) (.Values.lifecycle.preStop)}}
+        lifecycle:
+          {{- if .Values.lifecycle.postStart }}
+          postStart:
+            exec:
+              command: {{- range .Values.lifecycle.postStart }}
+                - {{ . }}
+              {{- end }}
+          {{- end }}
+          {{- if .Values.lifecycle.preStop }}
+          preStop:
+            exec:
+              command: {{- range .Values.lifecycle.postStart }}
+                - {{ . }}
+              {{- end }}
+          {{- end }}
+        {{- end }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
         {{- if .Values.startupProbe.enabled }}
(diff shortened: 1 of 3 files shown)


charts/minecraft/templates/deployment.yaml
```diff
@@ -142,6 +142,24 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         tty: true
         stdin: true
+        # Working here
```
itzg, line 145: Left over comment?

charts/minecraft/values.yaml
```diff
@@ -18,6 +18,10 @@ resources:
     memory: 512Mi
     cpu: 500m
 
+lifecycle:
+  postStart: {}
+  preStop: {}
```
itzg, line 23: Shouldn't these be lists?

```suggestion
  postStart: []
  preStop: []
```

itzg (commented):
Thanks! Please also bump the minor version of the chart. When the PR validation runs it'll check for that.

charts/minecraft/templates/deployment.yaml
allanger: Aye, sorry. It should be gone now

charts/minecraft/values.yaml
allanger: Sure they should. Now it must be fixed

allanger:
> Thanks! Please also bump the minor version of the chart. When the PR validation runs it'll check for that.

Ouch, I'm sorry for some dumb mistakes 🙃
I hope now it looks better

*As I see, you're using Squash Merge, so I've created a lot of commits. If you'd better see them rebased and squashed into one, I will do*

itzg (approved):

itzg:
No problem at all! And no need to squash them since like you said, I'll use a squash merge anyway.

PR #160, opened 2023-07-08
Added labels recomended in k8s docs

eaglesemanation:
I wanted to write a ServiceMonitor that would collect metrics from all k8s services with matching labels, and I found out that there are no common labels, so I could select all instances of this chart. Added some of recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,21 +1,21 @@
 apiVersion: v1
 name: minecraft
-version: 4.9.1
+version: 4.9.2
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
 keywords:
-- game
-- server
+  - game
+  - server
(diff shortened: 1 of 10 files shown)


itzg (approved):
Great idea

ishioni:
@eaglesemanation @itzg This PR broke (among other things) minecraft/templates/extraports-svc.yaml. This is because those resources are done in a range call and need to reference $ for variables (for example check https://github.com/itzg/minecraft-server-charts/blob/master/charts/minecraft/templates/extraports-ing.yaml#L16 against the newly added https://github.com/itzg/minecraft-server-charts/blob/master/charts/minecraft/templates/extraports-ing.yaml#L19 - the new additions compute to {} now)

itzg:
Thanks for the follow up and PR.

PR #162, opened 2023-07-09
fix missing global in range

ishioni:


eaglesemanation:
Thank you for this, should've done more testing before submitting PR

itzg (approved):
Thanks!

Issue #84, opened 2021-07-17
Should Minecraft be a StatefulSet?


ChipWolf:
StatefulSet seems more suitable than a Deployment for this type of workload, what are your thoughts?

itzg:
I prefer [using statefulsets](https://github.com/itzg/docker-minecraft-server/blob/master/examples/k8s/using-statefulset.yml), but changing these charts from deployments seems very disruptive for little to no functional gain.

ChipWolf:
The main reason I ask is we occasionally have an issue where a pod is stuck terminating & another one comes up to replace it, subsequently failing as it can't mount to the volume. I imagine StatefulSet wouldn't allow another pod to come up until the previous one was gone.

itzg:
STS's [rolling update strategy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates) seems to be advertised as similar behavior to deployment's recreate strategy. Go ahead and experiment with a statefulset in your scenario and see what happens.

itzg:
@billimek it feels like a similar conversation happened on the original helm repo, so I'm not able to find it there. What's your latest thoughts on the deployment/statefulset approaches and challenges.

billimek:
> @billimek it feels like a similar conversation happened on the original helm repo, so I'm not able to find it there. What's your latest thoughts on the deployment/statefulset approaches and challenges.

Personally speaking, I've gone full-circle on the idea that things should be `statefulsets` vs `deployments` meaning I believe that workloads like minecraft are appropriate for `deployments` and not `statefulsets`.  For the `minecraft-server` chart, it's most likely going to require persistence and using a `Recreate` strategy vs `Rollout` _should_ alleviate issues that @ChipWolf describe with a race condition of the new pod trying to consume the storage before the old pod terminates.  I personally have not seen this issue and I run several minecraft chart instances and wonder if there is a better way to go about remediating the problem that doesn't involve using `StatefulSet`.

Anecdotally, it seems some folks have some issues with StatefulSet workloads being too 'sticky' to a given host and making updates challenging.

ChipWolf:
> 
> 
> > @billimek it feels like a similar conversation happened on the original helm repo, so I'm not able to find it there. What's your latest thoughts on the deployment/statefulset approaches and challenges.
> 
[...]

PR #163, opened 2023-07-09
Option to deploy server as StatefulSet

FroznMine:
Some time ago #84 proposed to have the Minecraft servers deployed as StatefulSets. Due to possible problems with servers becoming too 'sticky' the idea was discarded. As I want to use this chart in conjunction with the auto-scale-up of [mc-router](https://github.com/itzg/mc-router), I need to deploy as a StatefulSet.

This PR adds `workloadAsStatefulSet` as a flag to deploy the servers (Java/Bedrock) as Deployment or StatefulSet. When deploying as a StatefulSet with persistence enabled, a PVC template is configured using the information that would have been used to create a single PVC in case of a Deployment.

### DIFF
--- a/charts/minecraft-bedrock/templates/datadir-pvc.yaml
+++ b/charts/minecraft-bedrock/templates/datadir-pvc.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.persistence.dataDir.enabled (not .Values.persistence.dataDir.existingClaim ) -}}
+{{- if and .Values.persistence.dataDir.enabled (not .Values.persistence.dataDir.existingClaim ) (not .Values.workloadAsStatefulSet) -}}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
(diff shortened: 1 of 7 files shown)


itzg (commented):
Thanks! It's very nice to have statefulset as an option. It'll just need the minor/middle chart version bumped for both. 

Just to double check, have you tested out both enabled/disabled modes?

danielr1996:
Any progress on this? Is there only the version that needs to be bumped or are there more things to do? 

Also I was wondering if there are reasons to choose a deployment over a statefulset, or if it would be fine to only ever use a statefulset because this would mean a lot less to maintain.

itzg:
Yes it looks like just the version bump and confirmation of testing enabled and disabled.

danielr1996:
I've opened a new PR https://github.com/itzg/minecraft-server-charts/pull/172/ with the bumped version, can verify that it creates a deployment or statefulset, depending on the setting.

PR #170, opened 2023-09-01
Added extraDeploy array

grantbruneau:
Adds extraDeploy array which allows us to include additional objects at deployment. My reason for this is to use helm pre-install/post-delete hooks which can automatically add/remove DNS entries for my minecraft servers.

[Reference](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks)

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,6 +1,7 @@
 name: Lint and Test Charts
 
-on: pull_request
+on: [pull_request, workflow_dispatch]
+
 
 jobs:
   lint-test:
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,6 +1,6 @@
 name: Lint and Test Charts
 
-on: [pull_request, workflow_dispatch]
+on: pull_request
 
 
 jobs:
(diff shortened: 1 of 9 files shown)


itzg (approved):
Thanks!

itzg:
The lint build failure probably isn't related to your change, but I'm not sure yet why it fails with

```
curl: (6) getaddrinfo() thread failed to start
```

grantbruneau:
> The lint build failure probably isn't related to your change, but I'm not sure yet why it fails with
> 
> ```
> curl: (6) getaddrinfo() thread failed to start
> ```

Looks like there was an error downloading bedrock at that moment. Can we retry the lint? I confirmed that URL works.
ERROR failed to download from https://minecraft.azureedge.net/bin-linux/bedrock-server-1.20.15.01.zip

itzg:
Had already re-run, but was waiting to try again today since it might have been a github workflow networking issue.

itzg:
Still failing in the same way. I really don't think it's related to your changes, so I'll experiment with a PR with no changes from master.

grantbruneau:
Seems like it might be the github runner, has anything changed on that end? My googling says maybe firewall issue.

grantbruneau:
@itzg I have it working on my fork. I updated the versions of KIND and metallb. Prior to that I had the same errors.

[reference](https://github.com/grantbruneau/minecraft-server-charts/actions/runs/6059767974/job/16443223144?pr=1)

itzg:
Thanks! Running it now

itzg:
Awesome, that fixed it. Thanks for digging into that.

grantbruneau:
Happy to help!

PR #173, opened 2023-09-26
Added minecraftServer.serverPort value to allow the pod to run on a different port for Bedrock

c-gunther:
The ping function doesn't work if the server is accessed on an external port (eg nodePort) that isn't 19132. This change allows the port of the pod to match the external port.


itzg (commented):
Thanks for adding that. You'll just need to bump the chart's minor version.

itzg (approved):

PR #177, opened 2023-11-02
Add MC Router chart

Bibz87:
This will enable MC Router deployment using Helm.

Chart was built using MC Router's regular YAML example file and other existing charts.


charts/mc-router/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            {{- include "mc-router.envBoolMap" (list "IN_KUBE_CLUSTER" "true") | nindent 12 }}
```
itzg, line 49: The `nindent 12` reintroduces Helm "noise" that I purposely wanted to avoid when creating the `env*` helpers. Can you put the indentation into those and get these to look like

```suggestion
{{- include "mc-router.envBoolMap" list "IN_KUBE_CLUSTER" "true" }}
```

charts/mc-router/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+              {{- include "mc-router.envBoolMap" (list "AUTO_SCALE_UP" .autoScaleUp.enabled) | nindent 12 }}
+              {{- include "mc-router.envMap" (list "CONNECTION_RATE_LIMIT" .connectionRateLimit) | nindent 12 }}
+              {{- include "mc-router.envMap" (list "CPU_PROFILE" .cpuProfilePath) | nindent 12 }}
+              {{- include "mc-router.envBoolMap" (list "DEBUG" .debug.enabled) | nindent 12 }}
+
+              {{- with .defaultServer }}
+                {{- include "mc-router.envMap" (list "DEFAULT" (printf "%s:%d" .host (.port | int))) | nindent 12 }}
+              {{- end }}
+
+              {{- include "mc-router.envMultilineMap" (list "MAPPING" (include "mc-router.formatMappings" .mappings)) | nindent 12 }}
```
itzg, line 63: Just wanted to confirm that this formats into nothing without any explicit mappings declared?

charts/mc-router/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+                    {{- end }}
+                  {{- end }}
+                {{- end }}
+              {{- end }}
+
+              {{- include "mc-router.envSecretMap" (list "NGROK_TOKEN" .ngrokToken.existingSecret .ngrokToken.tokenKey ) | nindent 12 }}
+              {{- include "mc-router.envBoolMap" (list "SIMPLIFY_SRV" .simplifySrv) | nindent 12 }}
+              {{- include "mc-router.envBoolMap" (list "USE_PROXY_PROTOCOL" .useProxyProtocol) | nindent 12 }}
+              {{- include "mc-router.envBoolMap" (list "VERSION" .showVersion) | nindent 12 }}
+            {{- end }}
```
itzg, line 87: Can you include an "extra env" section like

https://github.com/itzg/minecraft-server-charts/blob/1e56e34e883b53b2eac145c75df4971d4c2d71a0/charts/minecraft/templates/deployment.yaml#L363-L374

which serves as a nice escape hatch when new env vars are introduced before the chart is updated, if at all.

charts/mc-router/templates/service.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.apiPort }}
+      {{- if list "LoadBalancer" "NodePort" | has .Values.service.type }}
+      nodePort: {{ .Values.service.apiPort }}
+      {{- end }}
+      targetPort: api
+      protocol: TCP
+      name: api
```
itzg, line 16: It would be good to allow for API port to be optionally exposed. In fact, the API might be need to be declared in a separate service object since they may want to use `type: ClusterIP` for the API to keep it internal, but `type: NodePort` for the main routing port to expose externally.

charts/mc-router/values.schema.json
itzg, line 1: Thanks so much for getting ahead of this. It was way late in the game for the other charts to cover much at all in their schema.

itzg (changes requested):
Aweome, thanks! Just a couple of things and it's good to go.

charts/mc-router/values.schema.json
Bibz87: Yeah I like the fact that schemas give an extra layer of data validation :)

charts/mc-router/templates/deployment.yaml
Bibz87: Yes, this is what's generated with the default values:
``` yaml
env:
  - name: IN_KUBE_CLUSTER
    value: "true"
  - name: API_BINDING
    value: ":25580"
  - name: PORT
    value: "25565"
  - name: AUTO_SCALE_UP
    value: "false"
  - name: CONNECTION_RATE_LIMIT
    value: "1"
  - name: DEBUG
    value: "false"
  - name: METRICS_BACKEND
    value: "discard"
  - name: SIMPLIFY_SRV
    value: "false"
  - name: USE_PROXY_PROTOCOL
    value: "false"
  - name: VERSION
    value: "false"
```

charts/mc-router/templates/service.yaml
Bibz87: Done!

charts/mc-router/templates/deployment.yaml
Bibz87: Done!

itzg (approved):
Thanks for the quick follow up changes.

itzg:
Unfortunately, I'm not familiar with how the chart testing tool works or what it expects. I'm sure the chart itself is fine, but would be nice to get the test automation happy:

https://github.com/itzg/minecraft-server-charts/actions/runs/6734798515/job/18314217048?pr=177

Bibz87:
> Unfortunately, I'm not familiar with how the chart testing tool works or what it expects. I'm sure the chart itself is fine, but would be nice to get the test automation happy:
> 
> https://github.com/itzg/minecraft-server-charts/actions/runs/6734798515/job/18314217048?pr=177

I'll take a look at that. :)

itzg:
It looks like files like https://github.com/itzg/minecraft-server-charts/blob/master/charts/minecraft/ci/test-values.yaml get used as input to the test, but the default values should have been fine in this case.

Bibz87:
> It looks like files like https://github.com/itzg/minecraft-server-charts/blob/master/charts/minecraft/ci/test-values.yaml get used as input to the test, but the default values should have been fine in this case.

Yeah the problem's the default ports:

> Error: Service "mc-router-uhe6wa9e6h" is invalid: spec.ports[0].nodePort: Invalid value: 25565: provided port is not in the valid range. The range of valid ports is 30000-32767

itzg (approved):
Changes look good. Let's see what the chart test says.

itzg:
...it says: ✔️

PR #179, opened 2023-11-22
Add support for mc-backup using restic with rclone

Bibz87:
Also:
* Fixes a typo in the values' comments
* Adds documentation about a known immutable rclone configuration issue
* Add support for Restic environment variables to be read from Kubernetes resources


itzg (approved):
Thanks!

PR #182, opened 2023-12-22
Attemp to add mc-backup rsync method to helm

toddejohnson:
This is my WAG at adding mc-backup `rsync` support to the chart. It looks right. This is my first change to a helm chart. I'll have a few others to the proxy and perhaps support non-pvc based mounts for backup. I'm currently bypassing the chart and patching the yaml output before applying it. Would be nice to add the features to the chart and make life easier :)

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.12.0
+version: 4.13.0
 appVersion: SeeValues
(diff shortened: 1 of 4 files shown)


itzg (commented):
Looks good. You'll also need to bump the chart version -- let's go with a minor version bump. I'll kick off the verify build which will highlight that.

toddejohnson:
Version bumped. Yay!

itzg (approved):

PR #183, opened 2023-12-25
Expose DNS settings of deployment in values

toddejohnson:
I ran into a name resolution issue and the fix for me was [in ndots](https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html) as I pulled down the jar files. This exposes the DNS settings of the pod template so it can be altered in the chart if needed

```yaml
dnsPolicy: ClusterFirst
dnsConfig:
  options:
  - name: ndots
    value: '1'
```


itzg (dismissed):
Thanks!

itzg (commented):
Test build reports proxy chart also needs version bump.

itzg (approved):

PR #185, opened 2024-01-03
add subPath support for dataDir and backupDir

seanb4t:
Adding in support for specifying a subPath of a given volume to use for backup and dataDir ( partial support was already there for dataDir )


itzg (commented):
Awesome, thanks! Please also bump the chart's minor version and this will be good to go.

billimek:
It's probably also worthwhile adding a comment to the `values.yaml` file for the `Values.mcbackup.persistence.backupDir` section indicating that subPath is now an option.

itzg:
Agreed

PR #188, opened 2024-02-04
Add ability to change access modes

estenrye:
Enables the user to alter the access mode to define Jobs in the extraDeployment array that can mount and modify the contents of the data directory.


itzg (approved):
Thanks!

PR #189, opened 2024-02-14
Allow using existing secret for proxy rcon

toddejohnson:
TLDR: Get secret out of values.yaml into existing secret that can be out of git tree. Stole from Minecraft chart. 

I am redeploying my k3s cluster for the kids Minecraft using fluxcd so it will be quicker in the future. I found where I couldn't, that I know of, easily get the rcon secret out of the chart. I saw how it was done in the Minecraft chart and stole that and implemented it here.


itzg (approved):
Thanks for fixing that up.

PR #191, opened 2024-02-25
Configuration options for SSH access to Bedrock

tuxpeople:
Hi

This is related to https://github.com/itzg/docker-minecraft-bedrock-server/issues/86 and should allow users to set the configuration necessary for SSH access (eg. for https://github.com/Kaiede/Bedrockifier). 

I'm not an Helm pro, but I tried to test it as good as I could. It adds a few options to `Values.minecraftServer`:

- `enableSSH` wether or not enable SSH access. Defaults to `false`. If set to true, the SSH part will be added to the service and the env vaer `ENABLE_SSH` will be set.
- `nodePortSSH` is like the already existing `nodePort`
- `exposedPortSSH` allows to define the port exposed by the service. Defaults to `2222` if not set.
- `passwordSSH` allows to set the `RCON_PASSWORD` env variable to specify a password rather than to have it auto-generated

### DIFF
--- a/charts/minecraft-bedrock/Chart.yaml
+++ b/charts/minecraft-bedrock/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-bedrock
-version: 2.4.0
+version: 2.4.1
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server


tuxpeople:
There would also be the possibility to add a "enable backup" option to directly add and configure https://github.com/Kaiede/Bedrockifier as a sidecar container. What do you think?

charts/minecraft-bedrock/templates/minecraft-svc.yaml
```diff
@@ -39,5 +39,18 @@ spec:
     nodePort: {{ .Values.minecraftServer.nodePort }}
     {{- end }}
     protocol: UDP
+  {{- if .Values.minecraftServer.enableSSH }}
+  - name: minecraft-ssh
+    {{- if .Values.minecraftServer.exposedPortSSH }}
+    port: {{ .Values.minecraftServer.exposedPortSSH }}
+    {{- else }}
+    port: 2222
+    {{- end }}
```
itzg, line 48: Could use https://helm.sh/docs/chart_template_guide/function_list/#default here

itzg (commented):
Looks good. Just the minor Helm-ish suggestion.

charts/minecraft-bedrock/templates/minecraft-svc.yaml
tuxpeople: thanks!

itzg (commented):
Please also bump the minor version of the chart. Sorry, I forgot to check that in previous review.

itzg:
> There would also be the possibility to add a "enable backup" option to directly add and configure https://github.com/Kaiede/Bedrockifier as a sidecar container. What do you think?

I forgot to reply to this. This sounds like a great idea for a follow up PR.

PR #192, opened 2024-03-07
feat: allow setting annotations on mc-router services

d3adb5:
Allow setting annotations for the Service resources created by the `mc-router` chart in an independent fashion.

This is particularly useful when using services like MetalLB and setting either service type to LoadBalancer.

### DIFF
--- a/charts/mc-router/templates/api-service.yaml
+++ b/charts/mc-router/templates/api-service.yaml
@@ -5,6 +5,8 @@ metadata:
   name: {{ include "mc-router.fullname" . }}-api
   labels:
     {{- include "mc-router.labels" . | nindent 4 }}
+  annotations:
+    {{- toYaml .Values.services.api.annotations | nindent 4 }}
 spec:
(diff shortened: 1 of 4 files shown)


charts/mc-router/templates/api-service.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   name: {{ include "mc-router.fullname" . }}-api
   labels:
     {{- include "mc-router.labels" . | nindent 4 }}
+  annotations:
+    {{- toYaml .Values.services.api.annotations | nindent 4 }}
```
itzg, line 9: Does this render and deploy ok when empty?

itzg (commented):
Thanks!

charts/mc-router/templates/api-service.yaml
itzg: Thanks for checking

itzg (approved):

PR #196, opened 2024-04-11
Fix secret name for RCON password in proxy

bastibrunner:
Fixed secret reference name in deployment, there's no "-rcon" appended in the secret's template.

### DIFF
--- a/charts/minecraft-proxy/Chart.yaml
+++ b/charts/minecraft-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 3.6.0
+version: 3.6.1
 appVersion: SeeValues
 description: Minecraft proxy server (BungeeCord, Waterfall, Velocity, etc.)
 keywords:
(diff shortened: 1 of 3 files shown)


itzg:
@toddejohnson can you help review this change since you introduced the original with #189

toddejohnson (commented):
Sure thing! Not going to lie... I stole most that from:
https://github.com/itzg/minecraft-server-charts/blob/157686c9629b68320f13c7c7f8b3fc198caed7f3/charts/minecraft/templates/deployment.yaml#L100
https://github.com/itzg/minecraft-server-charts/blob/157686c9629b68320f13c7c7f8b3fc198caed7f3/charts/minecraft/templates/secrets.yaml#L1

In the secret I just wrapped it in the if to conditionally create it if it didn't exist. It does look like I forgot to pull over the  `-rcon`. Great catch @bastibrunner!

itzg:
Thanks for looking at it, @toddejohnson

itzg (approved):
Thanks! Let me know if you're ready for that to be merged.

toddejohnson:
Looks good to me!

PR #198, opened 2024-04-14
Fix replicaCount value

### LINKED ISSUE #195: Can't set `replicaCount` to 0
You are not able to set a `replicaCount` to 0. Doing so results in a `1` replica instead of `0` replicas. The use case is this is that I'd like to scale down the minecraft server while making other changes to the chart. However, when I apply the chart it keeps scaling it back to 1.

This is pretty easy to repro by running:
```
helm template   --set 'replicaCount=0' --set 'minecraftServer.eula=true' ./charts/minecraft | grep replicas
```

which returns
```
  replicas: 1
```

However using other values works:
```
helm template   --set 'replicaCount=3' --set 'minecraftServer.eula=true' ./charts/minecraft | grep replicas
```
returns
```
  replicas: 3
```

I believe this is caused by: https://github.com/helm/helm/issues/3164

That issue is attached to several work-around PRs, which may be helpful to resolve this.

David-Igou:
Fixes #195 

The default function is not necessary, and not recommended per the [helm documentation](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/)


itzg (commented):
Thanks for cleaning that up. Please also bump the chart's patch version.

PR #200, opened 2024-04-22
add customisable deployment and service lables

jurgenweber:
I need this for k3s serviiceLB, IMO it should be annotations but that is a different story. ;)

https://docs.k3s.io/networking/networking-services#creating-servicelb-node-pools

Thank you for your hard work!


itzg (commented):
Thanks! Please also bump the chart's minor version.

itzg (approved):

PR #202, opened 2024-04-23
add a restoreMode

jurgenweber:
this restoreMode allows you to get to the server files without the server running, so you can copy them off the PVC or copy something to the PVC. TUrn off restoreMode to bring back the server and you have your restored world.


itzg (approved):
Great idea!

PR #203, opened 2024-04-23
Side car

jurgenweber:
add the ability to add sidecars to the deployment; I made the asumption you will merge my other PR and bump'd accordingly.

### DIFF
--- a/charts/minecraft-bedrock/templates/deployment.yaml
+++ b/charts/minecraft-bedrock/templates/deployment.yaml
@@ -182,6 +182,9 @@ spec:
         {{- end }}
         securityContext:
           {{- toYaml .Values.securityContext | nindent 10 }}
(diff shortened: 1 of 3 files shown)


charts/minecraft-bedrock/templates/deployment.yaml
```diff
@@ -182,6 +182,9 @@ spec:
         {{- end }}
         securityContext:
           {{- toYaml .Values.securityContext | nindent 10 }}
+      {{- with .Values.extraContainers }}
```
itzg, line 185: Let's call it `sidecarContainers` to be consistent with Java edition chart

https://github.com/itzg/minecraft-server-charts/blob/54f91138c1504d77245be86aced7ebb23e8eb7c9/charts/minecraft/values.yaml#L101

PR #209, opened 2024-05-25
Allow configuration to mount extra volume to /data

jltobler:
I have an NFS share that contains the contents of my Minecraft server and would like to use it as the volume that gets mounted to `/data` directory in the container. As far as I can tell, it is not possible to configure the Minecraft Helm chart in this manner because an `emptyDir` volume is always mounted when the PVC is disabled and thus a conflict occurs.

To allow any defined extra volume to be mounted to the /data directory, I propose a new `extraDataVolume` configuration be introduced.

Thanks in advance for taking a look.

Example use with NFS:
```
persistence:
  extraDataVolume: nfs

extraVolumes:
    volumes:
      - name: nfs
        nfs:
          server: example.nfs.server
          path: /data/minecraft
```

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -399,6 +399,8 @@ envFrom: []
 persistence:
   labels: {}
   annotations: {}
+  ## specify an alternative volume to be mounted to /data instead of datadir.
+  # altDataVolumeName: ""
   ## minecraft data Persistent Volume Storage Class
   ## If defined, storageClassName: <storageClass>
   ## If set to "-", storageClassName: "", which disables dynamic provisioning
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.18.0
+version: 4.19.0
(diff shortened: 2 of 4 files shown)


charts/minecraft/values.yaml
```diff
@@ -399,6 +399,8 @@ envFrom: []
 persistence:
   labels: {}
   annotations: {}
+  ## specify an extra volume to be mounted to /data instead of datadir.
+  # extraDataVolume: ""
```
itzg, line 403: How about calling it "alternateDataVolumeName" (or "altDataVolumeName" to be shorter):

- "extra" implies "in addition to" but really this is replacing entirely the default data volume reference
- I propose adding the "Name" suffix to emphasize that is the name/reference to a pod volume and not the volume declaration in its entirety

itzg (commented):
Thanks for contributing this idea. I just have a suggestion about the value name.

charts/minecraft/values.yaml
jltobler: Great suggestion @itzg! I have updated the value name to `altDataVolumeName`. I agree that this name better reflects its intent. :+1: 

Thanks for the review!

itzg (approved):

PR #214, opened 2024-06-22
Fix: Make rclone config folder writable

FawenYo:
# Description

Since Kubernetes will make Secrets volume to be mounted as read-only, when mc-backup is trying to run `mkdir` in [this line](https://github.com/itzg/docker-mc-backup/blob/aea603f867e3aab2bb0165d7db93d595c6966ccb/scripts/opt/backup-loop.sh#L305), our pod will show the following error:

```
2024-06-22T00:59:20+0800 INFO sleeping 24h...
2024/06/22 00:59:24 ERROR : Failed to save config after 10 tries: failed to create temp file for new config: open /config/rclone/rclone.conf59648627: read-only file system
```

To fix this issue, we can create a new empty volume and use InitContainer to copy the content of the Secrets volume to the new volume. Then we can mount the new volume as read-write in the main container.

# Proven Image

I've tested this on my environment and now my `mc-backup` pod can successfully upload the backup to my drive using `rclone`.

image

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.19.0
+version: 4.20.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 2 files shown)


itzg (commented):
Sorry for the delay on this. 

Please bump the chart version and then this will be ready to merge.

FawenYo:
Hi @itzg, the Helm Chart version is bumped in the commit https://github.com/itzg/minecraft-server-charts/pull/214/commits/bb7dca3de9fdb3988819bda5f98f297c1f31661e

itzg:
I'm sorry, I'm not sure how I missed that.

PR #221, opened 2024-08-14
Add replicaCount configuration to minecraft-proxy chart

by77er:
I noticed that there's no way to set the number of replicas for the proxy deployment - just added the bare minimum, but can make additions as needed

### DIFF
--- a/charts/minecraft-proxy/Chart.yaml
+++ b/charts/minecraft-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 3.7.0
+version: 4.0.0
 appVersion: SeeValues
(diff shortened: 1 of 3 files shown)


charts/minecraft-proxy/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft-proxy
-version: 3.7.0
+version: 4.0.0
```
billimek, line 3: This looks good but I don't know if the nature of the change requires a major version bump.
 
```suggestion
version: 3.8.0
```

billimek (changes requested):

itzg (approved):

itzg:
@billimek github is blocking me from merging until you re-review. (I'm sure I could fix via admin settings, but that's too much work 😀)

billimek (approved):

PR #222, opened 2024-08-18
Allow optional definition of ingress

drewburr:
Removes the requirement of setting `ingress.enabled: false` for extraPorts. For example, an item in `extraPorts[]` needing no ingress currently requires:

```yaml
extraPorts:
  - name: map
    protocol: TCP
    ingress:
      enabled: false
    ...
```

If `ingress` is omitted, the following error is thrown:

```text
Error: template: minecraft-toiletbowl/charts/minecraft/templates/extraports-ing.yaml:3:26: executing "minecraft-toiletbowl/charts/minecraft/templates/extraports-ing.yaml" at <.ingress.enabled>: nil pointer evaluating interface {}.enabled
```

With this update, we can safely omit the `ingress` object entirely:

```yaml
extraPorts:
  - name: map
    protocol: TCP
    ...
```

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.20.0
+version: 4.20.1
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.20.1
+version: 4.21.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/templates/extraports-ing.yaml
+++ b/charts/minecraft/templates/extraports-ing.yaml
@@ -1,6 +1,6 @@
 {{- $minecraftFullname := include "minecraft.fullname" . }}
 {{- range .Values.minecraftServer.extraPorts }}
-{{- if default "" .ingress.enabled }}
+{{- if default "" (.ingress).enabled }}
 {{- $servicePort := .service.port }}
 {{- $serviceName := printf "%s-%s" $minecraftFullname .name }}
 ---


charts/minecraft/templates/extraports-ing.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $minecraftFullname := include "minecraft.fullname" . }}
 {{- range .Values.minecraftServer.extraPorts }}
-{{- if default "" .ingress.enabled }}
+{{- if default "" (.ingress).enabled }}
```
itzg, line 3: Very interesting. So this allows for nil-safe access to objects? So you happen to know where in the helm docs they describe that?

itzg (commented):
Thanks! Can you also bump the chart's minor version? (PR/merge process currently requires it 😞)

charts/minecraft/templates/extraports-ing.yaml
drewburr: I actually found this while poking through Google, with [this StackOverflow post](https://stackoverflow.com/a/77633748/19815719) leading me to the right answer. 

>In Helm we can wrap a value that we might consider optional in brackets.
>
> `type: {{ (.Values.fpm.service).type | default "NodePort" }}`
>
>Now when we try to run a helm template / install, it will consider .Values.fpm.service to be optional. We would still run into a Nil pointer if .Values.fpm is not set. So you must ensure to encapsulate up until where should be considered optional, as you can see in the example I provided, I did not place brackets around .type otherwise, it would try to find service which may not exist.

Digging into this further, [another StackOverflow post](https://stackoverflow.com/a/68807258/19815719) notes that this behavior comes from Go's `text/template` library. It seems that something like `{{ if default "" .ingress.enabled }}` is called a "pipeline" and adding the parenthesis makes a pipeline within the pipeline, like nested operators `if default "" (.ingress).enabled`. The nested pipeline returns a nil value under a suite of negative circumstances, which is then used instead of a nil pointer when accessing `.enabled` ([docs](https://pkg.go.dev/text/template#hdr-Actions)).

> If the value of the pipeline is empty, no output is generated
> ...
> The empty values are false, 0, any nil pointer or interface value, and any array, slice, map, or string of length zero.

Very interesting!

drewburr:
> Thanks! Can you also bump the chart's minor version? (PR/merge process currently requires it 😞)

Good catch! Thanks for the heads up, I'll keep this in mind in the future

charts/minecraft/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.20.0
+version: 4.20.1
```
itzg, line 3: Oops, that's actually the "patch" version. Can you bump it to "4.21.0"?

drewburr: Sure can, sorry about that!

itzg (approved):

PR #224, opened 2024-08-23
Reorder minecraft chart containers 

### LINKED ISSUE #223: Feature: Setting default container for minecraft chart
When using the `minecraft` chart with `.mcbackup.enabled`, the default container for commands such as `exec` and `logs` becomes the backup container, rather than the Minecraft server itself. This can sometimes made administration tedious, as I rarely need to interact with the backup container.

Kubernetes provides the [kubectl.kubernetes.io/default-container](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container) annotation to control this behavior, which we could choose to set to the server container instead.

I wanted to open this up to see if this was a feature we'd be interested in, or understand if this choice was deliberate. Pease see [this diff](https://github.com/itzg/minecraft-server-charts/compare/master...drewburr:minecraft-server-charts:patch-2?expand=1#diff-154ddbe694f97b6621716fbaded29f7989f94d41309adcb5807b30b6f19d1f38R51) for the changes I'm considering.

drewburr:
Updates the arrangement containers to ensure the minecraft server is always used for `exec` and `log`

Resolves #223

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.21.0
+version: 4.22.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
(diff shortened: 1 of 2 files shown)


itzg (approved):
Thanks for the follow up

PR #232, opened 2024-09-22
updated doc in Server Type to AUTO_CURSEFORGE

3deep5me:
CURSEFORGE is deprecated

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -171,7 +171,7 @@ minecraftServer:
   eula: "FALSE"
   # One of: LATEST, SNAPSHOT, or a specific version (ie: "1.7.9").
   version: "LATEST"
-  # This can be one of "VANILLA", "FORGE", "SPIGOT", "BUKKIT", "PAPER", "FTBA", "SPONGEVANILLA", "CURSEFORGE"
+  # This can be one of "VANILLA", "FORGE", "SPIGOT", "BUKKIT", "PAPER", "FTBA", "SPONGEVANILLA", "AUTO_CURSEFORGE"
   type: "VANILLA"
   # If type is set to FORGE, this sets the version; this is ignored if forgeInstallerUrl is set
   forgeVersion:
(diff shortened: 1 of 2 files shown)


charts/minecraft/values.yaml
```diff
@@ -171,7 +171,7 @@ minecraftServer:
   eula: "FALSE"
   # One of: LATEST, SNAPSHOT, or a specific version (ie: "1.7.9").
   version: "LATEST"
-  # This can be one of "VANILLA", "FORGE", "SPIGOT", "BUKKIT", "PAPER", "FTBA", "SPONGEVANILLA", "CURSEFORGE"
+  # This can be one of "VANILLA", "FORGE", "SPIGOT", "BUKKIT", "PAPER", "FTBA", "SPONGEVANILLA", "AUTO_CURSEFORGE"
```
itzg, line 174: Perhaps the rather than list them off exhaustively a second line of comment could point users to https://docker-minecraft-server.readthedocs.io/en/latest/types-and-platforms/ for the full set of options. For example `MODRINTH` also would need to be listed.

itzg (commented):
Thanks! Let me know what you think of the suggestion. 

Also, I know it's just a docs change but I believe the existing build checker will insist on a chart bump. Just the right most patch version will do.

charts/minecraft/values.yaml
3deep5me: I just added the change because i was irritated by using the CURSEFORGE/AUTO_CURSEFORGE.
I also added the link.
Should i update the chart version?
Feel free to commit on the branch.

itzg: Yep, that's totally fair

Please change the link to be the one I provided. The one you used is to a specific sub-section.

Yes, chart version still needs to be bumped. Sorry...I inherited the build setup as it is.

3deep5me: np. i like strict versioning :) 
Now it should be right?

itzg (approved):
Thanks!

Issue #165, opened 2023-08-15
How to configure velocity


arno4000:
According to https://github.com/itzg/minecraft-server-charts/blob/master/charts/minecraft-proxy/values.yaml#L127 the minecraft-proxy chart currently only supports config files, which are yaml. But velocity uses toml for its config file and velocity is supported by the chart. 

So my question is, how can I specify the config for velocity in the values.yaml of the minecraft-proxy chart?

itzg:
Yeah it seems to do extraneous processing to take the text block and turn it into yaml

https://github.com/itzg/minecraft-server-charts/blob/2406a8c2c6c075cf11bde390e8ac8ea006b49e4a/charts/minecraft-proxy/templates/configmap.yaml#L12

Removing the toYaml would probably fix it. 

Since I'm not personally working on this repo anymore, then a PR would be great.

arno4000:
It actually seems that the toYaml is not the problem. Instead, the `itzg/bungeecord` image seems to have a bug when mounting the `velocity.toml` to /server. I created a issue for this problem at https://github.com/itzg/docker-bungeecord/issues/133.

skylerspaeth:
I know this is an old issue, but in case it helps anyone else in the future, thanks to @ChipWolf's comment [here](https://github.com/itzg/minecraft-server-charts/pull/65#discussion_r608180401), I was able to configure velocity through the chart's `values.yaml`.

> ```yaml
> configFilePath: /server/velocity.toml
> config: |-
>  config-version = "1.0"
>  bind = "0.0.0.0:25577"
>  motd = "Look PR 65 - it works!"
>  show-max-players = 500
>  ...
> ```

And for what it's worth, I'm hoping to get [this PR ](https://github.com/itzg/minecraft-server-charts/pull/236)merged to add a comment with this example to the `values.yaml` (in addition to Velocity forwarding secret support).

PR #236, opened 2024-10-28
Allow velocity forwarding secret to be passed as value

skylerspaeth:
## Description
This repo indicates support for running a Velocity proxy, but without my proposed changes, I cannot spin one up when `player-info-forwarding-mode` is set to `modern` (which I have heard is the only _safe_ way to run Velocity).

This is due to the file at the default `forwarding-secret-file` path being missing. Since there is no apparent way to provide one without manual customizations, I figured I'd PR the changes I made back to master.

I'm coming into this project completely new, so forgive me if things are done differently than y'all usually do. Happy to make changes if needed. I used the _config_ volume mount (just above my new one) as an example:
```yaml
- name: config
  mountPath: {{ .Values.minecraftProxy.configFilePath }}
  subPath: config.yml
```
Hence why my` subPath` and `path` under the `volumes` section are hardcoded too (to `forwarding.secret`). That said, I assume the implication of doing so is that no matter what the user provides as `velocityForwardingSecretFilePath`, the _filename_ is going to remain the same, even though its location may change. I'm willing to adjust if needed.

Additionally, I added the example provided in this PR over 3 years ago to the `values.yaml` file, which will hopefully save others time trying to figure out how to configure Velocity with this chart: https://github.com/itzg/minecraft-server-charts/pull/65#discussion_r608180401

Thank you @ChipWolf for that info! 🙏

## Testing
I ensured both the changes worked together: example Velocity config along with forwarding changes. I couldn't test them individually because, at least in my case and with my knowledge, both are needed to prevent the pod from crashing.

To test them together, I used the following `values.yaml`, which contains the exact examples added to the upstream values file:
```yaml
minecraftProxy:
  type: VELOCITY
  velocityVersion: 3.4.0-SNAPSHOT
  serviceType: NodePort
  nodePort: 30565
  velocityForwardingSecret: "CHANGEME!"
  velocityForwardingSecretFilePath: /server/forwarding.secret
  configFilePath: /server/velocity.toml
  config: |-
    config-version = "2.7"
    bind = "0.0.0.0:25577"
    motd = "Hello from Helmified Velocity!"
    show-max-players = 500
    player-info-forwarding-mode = "modern"
```
The proxy shows up as expected in-game:

image

Even though I didn't mess with anything related to it, I also briefly tested the chart in the default BungeeCord mode just in case, and the MOTD appropriately changed back to the BungeeCord one in my server list.

I did not test these changes on any other sever types but I can't imagine there's any impact.

## Related Links
This is a list of helpful PRs and links that I referenced when making this PR (mostly here for my own future reference):
- https://github.com/itzg/minecraft-server-charts/pull/65
- https://github.com/itzg/minecraft-server-charts/issues/165
- https://github.com/itzg/docker-mc-proxy/issues/133


itzg (approved):
Thanks! That one was long awaited it seems.

Issue #227, opened 2024-08-26
Spawn protection setting bugged


alexyao2015:
The spawn protection config setting cannot be set to "0" as mentioned in the docs as it is currently bugged. Setting it to 0 will cause the envmap to evaluate to false and instead exclude the `SPAWN_PROTECTION` environment variable rather than including it as intended.

https://github.com/itzg/minecraft-server-charts/blob/ba5b3ba7168c3ac6f416e9c00aea35d37c0d9875/charts/minecraft/templates/_helpers.tpl#L28
https://github.com/itzg/minecraft-server-charts/blob/ba5b3ba7168c3ac6f416e9c00aea35d37c0d9875/charts/minecraft/templates/deployment.yaml#L174C44-L174C49

itzg:
Probably need to introduce an envIntMap similar to the envBoolMap template function.

thomascizeron:
Hi, 
I think it would be easier to modify the if statement in the `minecraft.envMap` template function to check if the value exists or the value is 0, allowing a value of 0 to render the template.
This way, you can use the envMap template function to render integer values as well. 
Something like this:  
`{{- if or (index . 1) (and (int (index . 1 | toString)) (eq (int (index . 1 | toString)) 0)) }}`
Maybe there is a better way to do this, but i think it's simpler than creating a new template function for integer values.
Happy to make a PR if you think this is a good idea.

itzg:
I'm worried that still wouldn't work. A zero is just as "false-y" as the absence of a value, so I'm not sure how your snippet can differentiate and might coerce all absent values into zero integers.

thomascizeron:
Right, after some tests the snippet don't work because of string casting. But this one works:
`{{- if or (index . 1) (kindIs "float64" (index . 1)) }}`
Either the value is true-y or it's a number (float64) (including 0).

Tested with:
| Type | index . 1 | Render |
|---|---|---|
| No value | | false |
| Null value | null | false |
| Empty string | "" | false |
| Empty array | [] | false |
| Empty map | {} | false |
| Zero integer | 0 | true |
| Non-zero integer | 1 | true |
| String | "string" | true |
| Array | [1, 2] | true |
| Map | {key: value} | true |

thomascizeron:
yeah it could work to also resorb `envBoolMap` with an additional clause on the `or`

itzg:
A PR would be great if you wanted to switch it to that.

thomascizeron:
[...]

PR #241, opened 2024-11-21
Fix: Spawn protection setting bugged 

thomascizeron:
See Issue #227 

Also resorb mc-router envBoolMap, can revert this if not wanted.

### DIFF
--- a/charts/mc-router/Chart.yaml
+++ b/charts/mc-router/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mc-router
-version: 1.2.3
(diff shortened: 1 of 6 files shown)


itzg (commented):
Sorry for the delay reviewing this. This looks really good!

Please bump the chart versions to next patch (3rd position) and this will be ready to merge.

thomascizeron:
I've rebased the branch and re-bump the chart version because it has changed since

itzg:
I just wanted to note that I forgot about the special value "default" that needed to be taken into account by the variable mapping helper.

PR #242, opened 2024-12-06
Default image pull policy to Always

itzg:
Since the default tag is "latest" and features arrive fairly frequently, it's more intuitive to have the latest image be more aggressive about checking for updates in the registry. The download itself is skipped if the latest image is already on the node, so it's a minor network penalty at pod creation time.


itzg:
Would have helped https://github.com/itzg/docker-minecraft-server/issues/3179

PR #243, opened 2024-12-13
Add option to disable TTY

Maddin-M:
Option to disable TTY for Minecraft Java servers, as described [here](https://docker-minecraft-server.readthedocs.io/en/latest/configuration/misc-options/#interactive-and-color-console).

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -365,6 +365,9 @@ minecraftServer:
     ## Set the externalTrafficPolicy in the Service to either Cluster or Local
     # externalTrafficPolicy: Cluster
 
+  ## set this to false to not have colorized logs
+  tty: true
+
   extraPorts:
     []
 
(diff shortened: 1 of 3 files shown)


charts/minecraft/values.yaml
```diff
@@ -364,6 +364,8 @@ minecraftServer:
     # loadBalancerSourceRanges: []
     ## Set the externalTrafficPolicy in the Service to either Cluster or Local
     # externalTrafficPolicy: Cluster
+    ## set this to false to not have colorized logs
+    tty: true
```
itzg, line 368: Can you move it up a level since it doesn't relate to rcon?

itzg (commented):
Thanks for the followup

charts/minecraft/values.yaml
Maddin-M: Sure, done

itzg (approved):
Thanks!

PR #250, opened 2025-03-02
Add support for MODRINTH_DOWNLOAD_DEPENDENCIES

2br-2b:
Once again, thank you so much for this project!

Right now, when trying to download Modrinth plugins, there's no support for [MODRINTH_DOWNLOAD_DEPENDENCIES](https://docker-minecraft-server.readthedocs.io/en/latest/mods-and-plugins/modrinth/#extra-options). This means that there is no way to specify to download required dependencies from Modrinth and they all must be manually specified. This PR fixes that

(Note - this is my first time modifying a helm chart, so you should probably make sure there's nothing else I missed changing. This should still provide backwards compatibility, but I don't know if you would also want to bump the version number)

### DIFF
--- a/charts/minecraft/templates/deployment.yaml
+++ b/charts/minecraft/templates/deployment.yaml
@@ -204,8 +204,8 @@ spec:
         {{- if .projects }}
 {{- template "minecraft.envMap" list "MODRINTH_PROJECTS" (join "," .projects) }}
         {{- end }}
-        {{- if .optionalDependencies }}
-{{- template "minecraft.envMap" list "MODRINTH_DOWNLOAD_OPTIONAL_DEPENDENCIES" .optionalDependencies }}
+        {{- if .downloadDependencies }}
+{{- template "minecraft.envMap" list "MODRINTH_DOWNLOAD_DEPENDENCIES" .downloadDependencies }}
         {{- end }}
 {{- template "minecraft.envMap" list "MODRINTH_ALLOWED_VERSION_TYPE" .allowedVersionType }}
     {{- end }}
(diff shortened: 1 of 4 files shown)


charts/minecraft/templates/deployment.yaml
```diff
@@ -204,7 +204,9 @@ spec:
         {{- if .projects }}
 {{- template "minecraft.envMap" list "MODRINTH_PROJECTS" (join "," .projects) }}
         {{- end }}
-        {{- if .optionalDependencies }}
+        {{- if .downloadDependencies }}
+{{- template "minecraft.envMap" list "MODRINTH_DOWNLOAD_DEPENDENCIES" .downloadDependencies }}
+        {{- else if .optionalDependencies }}
 {{- template "minecraft.envMap" list "MODRINTH_DOWNLOAD_OPTIONAL_DEPENDENCIES" .optionalDependencies }}
```
itzg, line 210: While you're in here can you remove all these references to MODRINTH_DOWNLOAD_OPTIONAL_DEPENDENCIES? That variable doesn't exist anymore.

itzg (commented):
Thanks! 

You'll need to bump the chart version in any case to make the helm deployment mechanics happy. I'd say bump the middle/minor version since it technically adds to the "interface".

itzg:
BTW there's always extraEnv for variables that the chart doesn't expose

https://github.com/itzg/minecraft-server-charts/blob/383793ab7e5579488f8f6e2f5efc5c1853b39b20/charts/minecraft/values.yaml#L421

In fact, it has become quite a maintenance headache to have mapped so many variables in the first place. My comment about the non-existent variable is a case in point 😀 .

2br-2b:
Should be all set now!

itzg (approved):
Changes look great. Just waiting on validation build.

PR #251, opened 2025-03-06
Fix user and group ID

blackliner:
I think the current values can't work, and they don't work when using nfs-csi:
```
/opt/bedrock-entry.sh: line 30: restify.err: Permission denied
```

### DIFF
--- a/charts/minecraft-bedrock/values.yaml
+++ b/charts/minecraft-bedrock/values.yaml
@@ -33,9 +33,9 @@ affinity: {}
 
 podSecurityContext:
   runAsUser: 1000
-  runAsGroup: 3000
+  runAsGroup: 1000
   runAsNonRoot: true
-  fsGroup: 2000
+  fsGroup: 1000
   seccompProfile:
     type: RuntimeDefault
 


itzg:
It's a security best practice. Refer to the contributing PR and issue

https://github.com/itzg/minecraft-server-charts/pull/155

Besides, those are just defaults. Set those values as you need during deploy time.

blackliner:
> It's a security best practice.
By "it" you mean using `securityContext:`, then yes, I fully agree, no changes there. But the chosen numbers for UID (1000) and GID (2000) won't work with fsgroup (3000). The example in the official docs does not specify these numbers for any meaningful reason, other than showing these are integers, like the usage of `foo`, `bar`, and `baz` that you might know of in many docs or tutorials.

itzg:
@uhthomas sorry to ping you after a 2 year old PR, but can you comment on the points brought up here?

itzg:
> Regarding defaults, shouldn't it be safe AND working by default?

This is the first I'm hearing that the values there don't work by default.

uhthomas:
The values shouldn't really matter, or even make a difference. They are all different to illustrate that they represent different things, and because it seems to be a good thing to do generally.

Changing these values all to 1000 is unlikely to help unless the underlying storage is already using group 1000, in which case the Helm values file can be configured.

I think the reasoning of keeping these values distinct is demonstrated quite well even here, the ID of the user, group and fs group are not related. A value of 1000 for all of them could incorrectly suggest correlation.

PR #253, opened 2025-03-15
feat: add nameOverride and fullnameOverride to minecraft chart

### LINKED ISSUE #252: Request: add `fullnameOverride` to minecraft Helm chart
I have a bunch of charts that are named things like `minecraft-<servername>` which creates pods called `minecraft-<servername>-minecraft`.

Additionally, I have some volsync automation and taskfiles that assume the PVC and the pod name will be the same (I create the PVC outside of the helm chart and pass it in).

Because of this mismatch, it causes occasional issues with my automation.

Would you be able to add a `fullnameOverride` to the chart so that I can pass in my name, and it will let me override the `-minecraft` append on the pod, service, etc?

You already have it in the mc-router chart.

joryirving:
Adds `nameOverride` and `fullnameOverride` to the chart/helpers, as per the [mc-router chart](https://github.com/itzg/minecraft-server-charts/blob/80e95abdf88492d5c1dee13860c10a704c14bcae/charts/mc-router/templates/_helpers.tpl#L13-L24).

Closes #252

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.24.0
+version: 4.25.0
 appVersion: SeeValues
 home: https://minecraft.net/
 description: Minecraft server
--- a/charts/minecraft/templates/_helpers.tpl
+++ b/charts/minecraft/templates/_helpers.tpl
@@ -11,9 +11,17 @@ Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "minecraft.fullname" -}}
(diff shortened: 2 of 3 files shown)


billimek:
Thanks @joryirving!  Can you please bump the chart version in `Chart.yaml` such that the new version is `4.25.0` given that this is a new feature?

billimek (approved):

janpfischer:
Since this change I get the following error message when trying to deploy the new chart version:
```
'Helm upgrade failed for release selfhosted/minecraft with chart minecraft@4.26.1:
      template: minecraft/templates/_helpers.tpl:32:14: executing "minecraft.fullname"
      at <.Values.fullnameOverride>: nil pointer evaluating interface {}.fullnameOverride'
```
I do not set a fullname in my values. I believe it has something to do with nameOverwrite and fullNameOverwrite are set to "" in values.yaml per default and the logic in the if-clauses.

Issue #254, opened 2025-03-15
Request: Upload Helm chart to OCI registry


joryirving:
With Helm v3.8.0, the OCI support became [GA](https://helm.sh/docs/topics/registries), which is an excellent chance to start publishing Helm charts to OCI-compliant registries. Quay / Dockerhub / GHCR supports OCI artifacts and a lot of projects have been using them to publish their helm charts as OCI artifacts.

It brings an opportunity to sign Helm charts stored as OCI Artifacts with [cosign](https://github.com/sigstore/cosign) to provide their integrity and use GitOps tooling such as Flux to reconcile them as they were stored as OCI artifacts. Flux can reconcile OCI Artifacts and verify their integrity before reconciling them.

itzg:
Is that something you can contribute to the repo as a PR or is there adjacent setup needed? I ask because I personally don't actively contribute to the charts these days.

joryirving:
I can take a shot at it, but it'll be a workflow, and will likely need some collaboration from a maintainer.
There's a prereq PR I'm submitting in a minute.

itzg:
I'm a maintainer for sure 😀 so I'll be glad to connect the dots.

joryirving:
Just wanted to see if anyone has had time to review the [PR](https://github.com/itzg/minecraft-server-charts/pull/256)

itzg:
Sorry, I have it queued up but haven't had time yet.

drag0n141:
For MC-Router is the new Release missing, because of that there is no OCI Helm Chart.

itzg:
That was quick 😀 The OCI chart publishing only got merged yesterday. 

I'll bump the other chart versions to force a publish.

itzg:
Now it's published

https://github.com/itzg/minecraft-server-charts/pkgs/container/minecraft-server-charts%2Fmc-router

PR #255, opened 2025-03-16
Fix charts.version for OCI chart rendering

joryirving:
This is a prerequisite for #254

```
failed to create resource: Secret "minecraft-<name>-rcon" is invalid: [metadata.labels: Invalid value: "minecraft-4.25.0+efa491f67fbd": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'), metadata.labels: Invalid value: "4.25.0+efa491f67fbd": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]
```

### DIFF
--- a/charts/minecraft-bedrock/templates/datadir-pvc.yaml
+++ b/charts/minecraft-bedrock/templates/datadir-pvc.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "minecraft.fullname" . }}
-    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}"
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
   {{- with .Values.persistence.labels  }}
--- a/charts/minecraft-bedrock/templates/datadir-pvc.yaml
+++ b/charts/minecraft-bedrock/templates/datadir-pvc.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "minecraft.fullname" . }}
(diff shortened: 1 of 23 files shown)


charts/minecraft-bedrock/templates/datadir-pvc.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "minecraft.fullname" . }}
-    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}"
```
itzg, line 9: I understand why the truncation and trim, but why just +’s are getting normalized to underscore?

joryirving: in OCI charts, it appends I believe... the SHA? So the `.Chart.Version` is `4.25.0+efa491f67fbd`, which is an invalid label value.

`a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character`

itzg: Yes I know about label validity rules, so clarifying the significance of plus sign makes sense now.

itzg, line 9: A template function would be good.  I know it doesn't help tremendously since it'll be per chart, but still would be nice to encapsulate the gory bits.

itzg (approved):
Thanks, that looks much better with the template function.

PR #263, opened 2025-04-10
fix: add catch for nil value

### LINKED ISSUE #262: Not providing fullNameOverride to minecraft chart is failing helm deployment
I tried to upgrade from 1.24 to 1.26. With 1.25 and https://github.com/itzg/minecraft-server-charts/issues/252 and https://github.com/itzg/minecraft-server-charts/pull/253 a feature was added to override the generated names for the resources. 

I have not specified a fullNameOverride in my values.yaml and am now getting the following error when trying to upgrade to 1.26:

```
'Helm upgrade failed for release selfhosted/minecraft with chart minecraft@4.26.1:
      template: minecraft/templates/_helpers.tpl:32:14: executing "minecraft.fullname"
      at <.Values.fullnameOverride>: nil pointer evaluating interface {}.fullnameOverride'
```

joryirving:
Add's guard for nil value for `nameOverride` and `fullnameOverride`

Fixes #262

### DIFF
--- a/charts/minecraft/templates/_helpers.tpl
+++ b/charts/minecraft/templates/_helpers.tpl
@@ -3,7 +3,8 @@
 Expand the name of the chart.
 */}}
 {{- define "minecraft.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- $nameOverride := (and .Values (get .Values "nameOverride")) }}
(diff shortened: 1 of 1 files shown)


itzg (commented):
Adding the guard seems like a decent solution, but it seems like the root cause must be something like a missing root/$ context anchor somewhere. Perhaps since you can't recreate the reported issue then it's finding a needle in the haystack?

joryirving:
Yes, without being able to reproduce it, all I have is the shared error message, and this might be a fix for it.

PR #264, opened 2025-04-12
fix: extraports ingress template context

mxtw:
this fixes the error caused by enabling ingress on an extraPort mentioned in #262


itzg (approved):
Yay! You found it. Thanks.

PR #270, opened 2025-05-04
Update mc-router chart for new auto scale options

### LINKED ISSUE #269: Router updates
Update `mc-router` chart to support new auto scale down and auto scale allow/deny list config options.

Sammcb:
Fixes #269

### DIFF
--- a/charts/mc-router/templates/clusterrole.yaml
+++ b/charts/mc-router/templates/clusterrole.yaml
@@ -8,7 +8,7 @@ rules:
   - apiGroups: [""]
     resources: ["services"]
     verbs: ["watch","list"]
-  {{- if .Values.minecraftRouter.autoScaleUp.enabled }}
+  {{- if .Values.minecraftRouter.autoScale.up.enabled }}
   - apiGroups: ["apps"]
     resources: ["statefulsets", "statefulsets/scale"]
     verbs: ["watch","list","get","update"]
--- a/charts/mc-router/templates/deployment.yaml
+++ b/charts/mc-router/templates/deployment.yaml
@@ -5,11 +5,10 @@ kind: Deployment
 metadata:
   name: {{ include "mc-router.fullname" . }}
   namespace: {{ .Release.Namespace }}
+  {{- with .Values.deploymentAnnotations }}
   annotations:
-    checksum/secret-config: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
-    {{- with .Values.deploymentAnnotations }}
     {{- toYaml . | nindent 4 }}
-    {{- end }}
+  {{- end }}
   labels:
     {{- include "mc-router.labels" . | nindent 4 }}
   {{- with .Values.deploymentLabels }}
@@ -26,9 +25,14 @@ spec:
     {{- .Values.deploymentStrategy | toYaml | nindent 4}}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- if or .Values.minecraftRouter.autoScale.allowDeny .Values.podAnnotations }}
       annotations:
+        {{- if .Values.minecraftRouter.autoScale.allowDeny }}
+        checksum/secret-config: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- end }}
+        {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
+        {{- end }}
       {{- end }}
       labels:
         {{- include "mc-router.labels" . | nindent 8 }}
@@ -142,10 +146,13 @@ spec:
               port: {{ $apiPort }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          {{- if or .Values.minecraftRouter.autoScale.allowDeny .Values.extraVolumes }}
           volumeMounts:
+            {{- if .Values.minecraftRouter.autoScale.allowDeny }}
             - name: secret-config
               mountPath: /etc/mc-router
               readOnly: true
+            {{- end }}
             {{- with .Values.extraVolumes }}
             {{- range . }}
             {{-   if .volumeMounts }}
@@ -153,6 +160,7 @@ spec:
             {{-   end }}
             {{- end }}
             {{- end }}
+          {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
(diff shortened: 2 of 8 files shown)


Sammcb:
@itzg This is ready for review! Absolutely no worries if you don't have time, I'm just excited to start using the new version once it's merged 😄

charts/mc-router/templates/clusterrole.yaml
```diff
@@ -8,7 +8,7 @@ rules:
   - apiGroups: [""]
     resources: ["services"]
     verbs: ["watch","list"]
-  {{- if .Values.minecraftRouter.autoScaleUp.enabled }}
+  {{- if .Values.minecraftRouter.autoScale.up.enabled }}
```
itzg, line 11: Unfortunately this would be a breaking change for existing chart users. You could do this as an or-expression to allow for both old and new value identifiers.

charts/mc-router/templates/deployment.yaml
```diff
@@ -25,9 +25,14 @@ spec:
     {{- .Values.deploymentStrategy | toYaml | nindent 4}}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- if or .Values.minecraftRouter.autoScale.allowDeny .Values.podAnnotations }}
       annotations:
+        {{- if .Values.minecraftRouter.autoScale.allowDeny }}
+        checksum/secret-config: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
```
itzg, line 31: I'm not sure from just this how the annotation name and secrets file relates to allow/deny list.

charts/mc-router/templates/deployment.yaml
```diff
@@ -54,7 +59,12 @@ spec:
 {{- include "mc-router.envMap" (list "PORT" $minecraftPort) }}
 
             {{- with .Values.minecraftRouter }}
-{{- include "mc-router.envMap" (list "AUTO_SCALE_UP" .autoScaleUp.enabled) }}
+{{- include "mc-router.envMap" (list "AUTO_SCALE_UP" .autoScale.up.enabled) }}
```
itzg, line 62: Same backward compatibility comment

charts/mc-router/templates/deployment.yaml
```diff
@@ -136,13 +146,20 @@ spec:
               port: {{ $apiPort }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
-          {{- with .Values.extraVolumes }}
+          {{- if or .Values.minecraftRouter.autoScale.allowDeny .Values.extraVolumes }}
           volumeMounts:
+            {{- if .Values.minecraftRouter.autoScale.allowDeny }}
+            - name: secret-config
```
itzg, line 152: and related to previous comment

charts/mc-router/templates/deployment.yaml
```diff
@@ -157,13 +174,20 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
 
-      {{- with .Values.extraVolumes }}
+      {{- if or .Values.minecraftRouter.autoScale.allowDeny .Values.extraVolumes }}
       volumes:
+        {{- if .Values.minecraftRouter.autoScale.allowDeny }}
+        - name: secret-config
```
itzg, line 180: Same

charts/mc-router/templates/secret.yaml
itzg, line 1: Maybe this helps answer my previous comments, however I'm not sure if that file needs to be a secret and I'm not sure if this is forcing too much convention on the  users.

charts/mc-router/values.schema.json
```diff
@@ -60,16 +60,44 @@
     "minecraftRouter": {
       "type": "object",
       "properties": {
-        "autoScaleUp": {
```
itzg, line 63: Regarding my backward compatibility commments, I'm not sure if the old value/object needs to be also included here. I'd have to look if JSON schema has a deprecated flag.

itzg (commented):
Sorry for the delay in reviewing this.

charts/mc-router/templates/deployment.yaml
Sammcb: This is done because for now, this secret only exists if the user specifies an auto scale allow/deny list. If the secret does exist, and the users makes changes to that list then the `mc-router` pod needs to be restarted to receive the changes which is what the annotation triggers. I could instead make this a TPL function called `secretEnabled` or something that simply wraps `.Values.minecraftRouter.autoScale.allowDeny` to better clarify intent?

charts/mc-router/templates/secret.yaml
Sammcb: I think it could either be a secret or config map. If you would prefer, I'm happy to make that an additional option in the values.

charts/mc-router/values.schema.json
Sammcb: (This applies to all other backwards compatible comments). I'm happy to make changes to allow for backwards compatibility, but wondering how you would feel about simply bumping the chart major version and leaving this as a breaking change? I completely understand if you would prefer to make it backwards compatible, just wanted to check first 😄

charts/mc-router/templates/deployment.yaml
itzg: Aha, I had missed the sha operation, so now I understand and agree with the "checksum/" prefix. Otherwise perhaps a more descriptive name like "checksum/allow-deny-secret" would make me all the way happy.

charts/mc-router/templates/secret.yaml
itzg: Yeah, if it doesn't cause too much of a mess with the values and template, it would be nice to give the option of secret or configmap. Otherwise, just as a secret now seems decent to me.

charts/mc-router/values.schema.json
itzg: Good point, that is supposed to be point of major versions 😀 Even a minor version bump is probably somewhat legit for just an interface change. 

How about this:
- keep it as a minor version bump
- don't worry about backward compatible handling
- could place somewhere a condition and `fail` if the old value is set. (I'm not sure the best convention of where to declare such template guards)

Sammcb:
Renamed some object to better clarify their purpose. Users can now choose to use a Secret or ConfigMap for storing allow/deny list config. Also, with the addition of the `"additionalProperties": false` value, if a user has `autoScaleUp` set, they will be warned that it does not match the schema.

itzg (approved):
Those updates looked great. Thanks!

PR #272, opened 2025-08-16
Add HTTPRoute support to minecraft chart

toddejohnson:
I've moved from ingress to gateway and as part of that now use httproute more. This attempts to add [HTTPRoute](https://gateway-api.sigs.k8s.io/api-types/httproute/) support to the extraports. 

Todo:
* [x] Bump chart version
* [x] Add example to values.yaml
* [ ] Validate values.schema.json with new items

### DIFF
--- a/charts/minecraft/Chart.yaml
+++ b/charts/minecraft/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: minecraft
-version: 4.26.4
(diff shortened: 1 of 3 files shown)


itzg:
Thanks. Yeah, adding to the values schema would be great. 

In general, I'm wondering when you're needing http for the Minecraft server? For map exposure mods/plugins?

PR #256, opened 2025-03-16
OCI chart release

joryirving:
Update the release pipeline to continue to push to the ghcr helm repository, as well as push an OCI image and sign it.

This also cleans up the chart.version reference using a template.
This publishes the objects as `ghcr.io/itzg/minecraft-server-charts/<chartname>`

Successful run: https://github.com/joryirving/minecraft-server-charts/actions/runs/13891040259/job/38862979875

OCI artifact for mc-router: https://github.com/joryirving/minecraft-server-charts/pkgs/container/minecraft-server-charts%2Fmc-router
OCI artifact for minecraft: https://github.com/joryirving/minecraft-server-charts/pkgs/container/minecraft-server-charts%2Fminecraft

I've tested it successfully in my own cluster, and the changes to chart-version work.

Unfortunately, I couldn't get cosign working:
https://github.com/joryirving/minecraft-server-charts/actions/runs/13890798966/job/38862394740#step:11:21

Closes #254

PR #273, opened 2025-09-06
Bump version for OCI push

ProfMoo:
I was unable to find `rcon-web-admin` in an OCI repository:

``` bash
> helm pull oci://ghcr.io/itzg/minecraft-server-charts/rcon-web-admin
Error: GET "https://ghcr.io/v2/itzg/minecraft-server-charts/rcon-web-admin/tags/list": response status code 404: name unknown: repository name not known to registry
```

Seems that even though [this PR](https://github.com/itzg/minecraft-server-charts/pull/256) added OCI support, the version for `rcon-web-admin` was never bumped to actually trigger an upload.

Hopefully this PR solves this.

### DIFF
--- a/charts/rcon-web-admin/Chart.yaml
+++ b/charts/rcon-web-admin/Chart.yaml
@@ -3,7 +3,7 @@ name: rcon-web-admin
 home: https://github.com/rcon-web-admin/rcon-web-admin
 description: RCon Web UI for managing game servers
 type: application
-version: 1.1.0
+version: 1.1.1
 appVersion: "0.14.1-1"
 keywords:
   - game


ProfMoo:
> The upstream rcon web admin is quite flaky. In fact, I had temporarily stopped support of the image. As such, I'm not highly motivated to debug the validation failure from the build.
> 
> Please investigate and let me know if there's anything you need me to do.

@itzg Yeh I started looking into this more and I see how old and outdated the `rcon-web-admin` project is. I think I'm simply going to remove it from my cluster 😄.

Thanks for the swift reply and for maintaining the other charts! I'll go ahead and close.

PR #279, opened 2026-01-21
fixes(#278): enable dualstack services

### LINKED ISSUE #278: Service doesn't allow to enable Dual Stack support
It's not possible to configure Dual Stack support for the Service (`ipFamilies` and `ipFamilyPolicy` property).

danielr1996:
Added support for configured ipFamily and ipFamilyPolicy. I think according to the docs

> Allocates both IPv4 and IPv6 cluster IPs for the Service when dual-stack is enabled. If dual-stack is not enabled or supported, it falls back to single-stack behavior.

defaulting to PreferDualStack should be fine and work in every case (IPv4 only, IPv6 only, dual stack), but we could also default to a more conservative IPv4 SingleStack policy.

fixes #278

### DIFF
--- a/charts/minecraft/values.yaml
+++ b/charts/minecraft/values.yaml
@@ -303,8 +303,8 @@ minecraftServer:
   ## Set the externalTrafficPolicy in the Service to either Cluster or Local
   # externalTrafficPolicy: Cluster
   externalIPs:
-  ipFamilies: [IPv4, IPv6]
-  ipFamilyPolicy: PreferDualStack
+  ipFamilies:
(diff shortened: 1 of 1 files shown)


billimek:
After looking at the [dual stack docs](https://kubernetes.io/docs/concepts/services-networking/dual-stack/) I think we should default to single stack and allow values to override this as a unit (i.e. it would set both `ipFamilyPolicy: PreferDualStack` and `ipFamilies:` together.)

I suspect the defaulting to enabled is what is causing the CI to fail, too.

billimek (approved):

PR #286, opened 2026-04-12
Fix websocket url load fails and added persistence 

### LINKED ISSUE #258: broken minecraft-rcon-web chart logic
https://github.com/itzg/minecraft-server-charts/blob/218074bbc6d4f48d19d3cb2bbd53f60b50be9950/charts/rcon-web-admin/templates/deployment.yaml#L75C1-L76C1

# Description

Basically there are several problems with this chart I've encountered; Firstly, the chart expects you to be using the `ingress.enabled` and using the in-built ingress manifest, however, if you're using other ingress controllers like istio, or API Gateway you will of course have this disabled which then leads you into other problems, where now it expects itself to reference the service as a `type: Loadbalancer`, this feels like something where you should be able to use custom ingress controllers without needing to then also layer a Loadbalancer IP requirement as well.

This also meant that the `RWA_WEBSOCKET_URL_SSL` and `RWA_WEBSOCKET_URL` env variables were completely wrong and needed to be patched.

I also noticed if you didn't use the `ingress.enabled` with `true` that it attempts to install `jq` however, it fails, firstly because the `securityContext` and `podSecurityContext` will cause it to break with the following errors:

```log
E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
```

Once you figure that out and resolve it by changing the settings, you end up hitting another issue in the container

```log
W: The repository 'http://security.debian.org/debian-security stretch/updates Release' does not have a Release file.
W: The repository 'http://deb.debian.org/debian stretch Release' does not have a Release file.
W: The repository 'http://deb.debian.org/debian stretch-updates Release' does not have a Release file.
E: Failed to fetch http://security.debian.org/debian-security/dists/stretch/updates/main/binary-amd64/Packages  404  Not Found [IP: 151.101.82.132 80]
E: Failed to fetch http://deb.debian.org/debian/dists/stretch/main/binary-amd64/Packages  404  Not Found [IP: 151.101.82.132 80]
E: Failed to fetch http://deb.debian.org/debian/dists/stretch-updates/main/binary-amd64/Packages  404  Not Found [IP: 151.101.82.132 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
/bin/sh: 12: jq: not found
```

Which basically I had to fix with something like follows which was a complete hack:

```bash
sed -i 's|deb.debian.org|archive.debian.org|g' /etc/apt/sources.list
sed -i 's|security.debian.org|archive.debian.org|g' /etc/apt/sources.list
sed -i '/stretch-updates/d' /etc/apt/sources.list
```

After all this I still couldn't get it to work correctly, I'm no expert but I think this chart / image needs some TLC and hope these insights help with potentially improving it moving forward 🙏

FernandoRod13:
# Fix issue #258 rcon-web-admin chart issues

## Problem Description

The rcon-web-admin chart has two critical issues that prevent it from working correctly in Kubernetes environments:

### Issue 1: Broken WebSocket URL Configuration

**Location**: Container startup script

**Problem**: The container's default ENTRYPOINT script attempts to:
1. Install `jq` via `apt-get` to parse Kubernetes API responses
2. Query the Kubernetes API to dynamically discover the LoadBalancer IP
3. Set `RWA_WEBSOCKET_URL` based on the discovered IP

This fails in most Kubernetes deployments because:
- `apt-get` may fail due to read-only root filesystems or permission issues
- Even if `jq` installation succeeds, the API call may return an empty response
- When `jq` fails, it outputs nothing, resulting in `RWA_WEBSOCKET_URL=ws://:4327` (empty host)

**Result**: The web UI loads but fails to connect to the backend websocket, showing "Invalid url for WebSocket ws://:4327"

### Issue 2: No Database Persistence

**Location**: `values.yaml` - `persistence` section

**Problem**: The chart uses an `emptyDir` volume for the database (`/opt/rcon-web-admin/db`), which means:
- All server configurations are lost on pod restart
- User settings are reset
- A new server ID is generated each restart, confusing clients

**Result**: Users must re-configure their Minecraft servers after every deployment update or pod restart.

### Fix 1: Add Persistent Storage Support

Add proper PVC-based persistence to the values.yaml:

```yaml
persistence:
  enabled: true
  storageClass: ""        # Uses default storage class
  accessMode: ReadWriteOnce
  size: 1Gi
```

This requires updating the deployment to use a PVC instead of emptyDir.

### Fix 2: Add WebSocket URL Environment Variables

Add environment variable support for websocket configuration with sensible defaults:

```yaml
rconWeb:
  websocketUrl: "ws://localhost:4327"      # Default for internal/LB access
  websocketUrlSsl: "wss://localhost:4327"  # Default for SSL
```

**Why `localhost:4327` works**: The browser connects to the LoadBalancer IP (e.g., `http://192.168.1.100:8080`), which forwards traffic to the pod. The pod's port 4327 is exposed via the same service, so `localhost:4327` resolves correctly when accessed through the LoadBalancer.

**When to override**: If using an external proxy or ingress with different hostnames, users can set custom URLs.

### Fix 3: Update Deployment for Persistence

The deployment needs to conditionally use a PVC when persistence is enabled:

```yaml
volumes:
  {{- if .Values.persistence.enabled }}
  - name: db
    persistentVolumeClaim:
      claimName: {{ include "rcon-web-admin.fullname" . }}-db
  {{- else }}
  - name: db
    emptyDir: {}
  {{- end }}
```

### Fix 4: Remove deployment command

```diff
-       {{- if not .Values.ingress.enabled }}
-          command:
-            - '/bin/sh'
-            - '-c'
-            - |-
-              # Installing jq to parse k8s response
-              export DEBIAN_FRONTEND=noninteractive
-              apt-get -qq update >/dev/null && apt-get -qq install -y jq > /dev/null
-              # Configuring k8s API auth
-              APISERVER=https://kubernetes.default.svc
-              SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
-              NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
-              TOKEN=$(cat ${SERVICEACCOUNT}/token)
-              CACERT=${SERVICEACCOUNT}/ca.crt
-              # Querying for websocket service
-              WS_SERVICE="$(curl --silent --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X GET ${APISERVER}/api/v1/namespaces/{{ .Release.Namespace }}/services/{{ include "rcon-web-admin.fullname" . }})"
-              {{- if contains "LoadBalancer" .Values.service.type }}
-              WS_IP="$(echo "$WS_SERVICE" | jq -r .status.loadBalancer.ingress[0].ip)"
-              WS_PORT="{{ .Values.service.wsPort }}"
-              {{- else }}
-              {{- fail "Selected service type is not supported"}}
-              {{- end }}
-              export RWA_WEBSOCKET_URL="ws://$WS_IP:$WS_PORT"
-              export RWA_WEBSOCKET_URL_SSL="wss://$WS_IP:$WS_PORT"
-             /usr/local/bin/node src/main.js start
-          {{- end}}
```

## Workaround for Existing Deployments

Users with existing deployments can work around these issues by:

1. **Patch the deployment with websocket URL** (using `localhost` works when accessed via LoadBalancer):
```bash
kubectl patch deployment rcon-web-rcon-web-admin \
  --namespace minecraft \
  --type=strategic \
  -p '{"spec":{"template":{"spec":{"containers":[{"name":"rcon-web-admin","env":[{"name":"RWA_WEBSOCKET_URL","value":"ws://localhost:4327"},{"name":"RWA_WEBSOCKET_URL_SSL","value":"wss://localhost:4327"}]}]}}}}'
```

2. **Add persistence**:
```bash
kubectl patch deployment rcon-web-rcon-web-admin \
  --namespace minecraft \
  --type=strategic \
  -p '{"spec":{"template":{"spec":{"volumes":[{"name":"db","persistentVolumeClaim":{"claimName":"rcon-web-db"}}]}}}}'
```

3. **Create a PVC**:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: rcon-web-db
spec:
  accessModes: [ReadWriteOnce]
  resources:
    requests:
      storage: 1Gi
```

## Testing

To verify the fixes:

1. Deploy with persistence enabled
2. Configure a Minecraft server in the web UI
3. Restart the deployment: `kubectl rollout restart deployment rcon-web-rcon-web-admin`
4. Verify server configuration persists after restart
5. Verify websocket connection works from browser

### DIFF
--- a/charts/rcon-web-admin/Chart.yaml
+++ b/charts/rcon-web-admin/Chart.yaml
@@ -3,7 +3,7 @@ name: rcon-web-admin
 home: https://github.com/rcon-web-admin/rcon-web-admin
 description: RCon Web UI for managing game servers
 type: application
-version: 1.1.0
+version: 1.2.0
 appVersion: "0.14.1-1"
 keywords:
   - game
--- a/charts/rcon-web-admin/ci/test-values.yaml
+++ b/charts/rcon-web-admin/ci/test-values.yaml
@@ -1,5 +1,7 @@
 service:
-  type: LoadBalancer
+  type: ClusterIP
 
 rconWeb:
   password: test
+  rconHost: ""
+  rconPort: 0
--- a/charts/rcon-web-admin/templates/NOTES.txt
+++ b/charts/rcon-web-admin/templates/NOTES.txt
@@ -1,3 +1,5 @@
+## Configuration
+
 1. Get the application URL by running these commands:
 {{- if .Values.ingress.enabled }}
 {{- with .Values.ingress }}
@@ -18,3 +20,15 @@
   echo "Visit http://127.0.0.1:8080 to use your application"
   kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:$CONTAINER_PORT
 {{- end }}
+2. The web UI websocket is configured to connect to localhost:4327 by default.
+   This works when accessing via LoadBalancer since the service forwards to the same pod.
+
+3. For persistence (recommended), ensure persistence.enabled=true in your values.
+
+## Troubleshooting
+
+If the websocket fails to connect, you may need to set custom websocket URLs:
+
+rconWeb:
+  websocketUrl: "ws://<your-loadbalancer-ip>:{{ add .Values.service.httpPort 1 }}"
+  websocketUrlSsl: "wss://<your-loadbalancer-ip>:{{ add .Values.service.httpPort 1 }}"
--- a/charts/rcon-web-admin/templates/db-pvc.yaml
+++ b/charts/rcon-web-admin/templates/db-pvc.yaml
@@ -1,0 +1,17 @@
+{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "rcon-web-admin.fullname" . }}-db
+  labels:
+    {{- include "rcon-web-admin.labels" . | nindent 4 }}
+spec:
+  accessModes:
+    - {{ .Values.persistence.accessMode | default "ReadWriteOnce" }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size | default "1Gi" }}
+  {{- if .Values.persistence.storageClass }}
+  storageClassName: {{ .Values.persistence.storageClass }}
+  {{- end }}
+{{- end }}
--- a/charts/rcon-web-admin/templates/deployment.yaml
+++ b/charts/rcon-web-admin/templates/deployment.yaml
@@ -26,8 +26,14 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       volumes:
-        - name: db
-          emptyDir: {}
+         {{- if .Values.persistence.enabled }}
+         - name: db
+           persistentVolumeClaim:
+             claimName: {{ include "rcon-web-admin.fullname" . }}-db
+         {{- else }}
+         - name: db
+           emptyDir: {}
+         {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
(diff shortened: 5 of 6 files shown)


itzg:
Thanks for working on this. The original logic was indeed over ambitious and I'm surprised I approved the use of apt during an entry point.

itzg (commented):
Looks great. Please also bump the chart version.

FernandoRod13:
Can't believe I missed it in my original commit. All done!

itzg:
As I mentioned in the linked issue, that rcon web admin (which is not mine) doesn't seem to work most of the time. I'm guessing that's why the test deploy is failing.

itzg (approved):
Ah yes, that fix makes total sense

FernandoRod13:
There are several issues on the CI side. I have tested this on my own cluster and the rcon-web is able to connect to the server. 


## Issues

### Metallb IP Pool Allocation

The default `rconHost` is set to `127.0.0.1` while the Metallb ip address-pool is  set to `192.168.1.240-192.168.1.250`. There is no overlap so no IP would be allocated making the chart fail to complete deployment. By setting the `service.type` to `ClusterIP` we can bypass Metallb. 

Alternatively we could set an IP for the `rconHost` within the IP pool but then there doesn't seem to be a Minecraft deployment ready to accept the required websocket connection so I think it would also fail.

### No Minecraft deployment to Test RCON Connection.

#### Options

1. Use `ClusterIP` + empty `rconHost` (simplest for CI)
2. Deploy a test Minecraft server in the test namespace for CI
3. Add a sidecar that listens on port `25575` as a mock RCON server

PR #290, opened 2026-05-09
Add port environment var passthrough when servicePort is enabled for LoadBalancer service

Azure-Owl:
When enabling servicePort the endpoint and listening port on the MC server inside the port, both values now receive their input from servicePort


itzg (commented):
Thanks! Please also bump the patch part of the chart version.

itzg (approved):

